PR #2, opened 2019-07-26
Background processing and spreadsheet support.

marc:
Add delayed_job for background processing.
Add iostreams and creek for Excel and CSV support.


marc:
Good point about large Excel files!
We can advise the users to use CSV files for large files since that format is also supported and does not require the full file to be loaded into memory.

PR #3, opened 2019-07-26
add SpawningSuccess model

bgarr:
### Description
Added model and migration for first Excel spreadsheet, enum for spawning_success
*@marc* we thought we should add a reference to the model you use to perform uploads, so if they need to roll back an upload, they can refer to the file whose records they'd like to remove from the db.

### Type of change

### How Has This Been Tested?
We ran the migration and added the model, constants for the headers


corneliusellen:
Looks great! bitmoji

PR #17, opened 2019-07-26
Add UntaggedAnimalAssessment model

glassjoseph:
### Description
Adds UntaggedAnimalAssessment model and migration


glassjoseph:
Resolves issue #5

glassjoseph:
Untagged sheet header is "measurement _data_", so we've copied that exactly. It's likely a typo, but we're still waiting for confirmation that we can change the header.

PR #74, opened 2019-10-11
Add annotate gem

wesleyeewong:
### Description
I've added this annotate gem: https://github.com/ctran/annotate_models

Purely to ease the development of this project.
I feel this would be a useful addition, as devs won't have to check migration files/or db itself to see what attributes the models have. And as evident from the failing tests, and incorrect attributes for some model factories.

Let me know if this needs more discussions :) I'm also open for suggestions.

### Type of change

* Development process improvements

### How Has This Been Tested?

Ran complete spec tests to make sure nothing else is broken. There's currently 1 failing test, that is fixed from my other PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ group :development do
   gem 'web-console', '>= 3.3.0'
   gem 'listen', '>= 3.0.5', '< 3.2'
 
+  gem 'annotate',                      require: false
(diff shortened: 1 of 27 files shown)


rafaltrojanowski:
Great initiative! I've also fixed this failing test in my PR 🍻

PR #76, opened 2019-10-25
Authentication

### LINKED ISSUE #71: Backend/Frontend - Authentication
We need basic authentication for at least one user. Login page.

glassjoseph:
Resolves #71 

### Description
Adds Devise gem authentication to app.

Requires authentication for accessing and uploading data. Users cannot sign up on their own and need to be added by a developer. Mailer will need to be configured in production if we want Devise emails for password reset, etc. 

### Type of change

* Documentation update


jcavena:
@glassjoseph I'd move the `before_action :authenticate_user!` out of the individual controllers and into the application controller. You can then use `skip_before_action :authenticate_user!` in the one controller/actions combo where you won't need it.

glassjoseph:
Thanks for the suggestion! I'll do that.

rafaltrojanowski:
@jcavena I'm not sure about this approach. It's DRY, yes, but won't be 'authenticate_user' method triggered for Devise controllers as well, ie. login, register and logout?

rafaltrojanowski:
@glassjoseph Thanks for checking it. I've got to hand it to you, @jcavena :)

jcavena:
@rafaltrojanowski Devise might do that automatically, like @glassjoseph found, but you can always `skip` callbacks in a particular controller for actions you don't want it to run on. Easier to skip the few exceptions than it is to remember to include in all new controllers down the line.

jcavena:
@glassjoseph Any chance you can take a look at the merge conflicts? These may have shown up after your changes so not your fault.

glassjoseph:
@jcavena Thanks, should be good to go now.

config/initializers/devise.rb
```diff
@@ -0,0 +1,299 @@
...
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
jtu0: @pollygee What email address should someone in Abalone use if they need to ask for Devise help? we should set the config to use that address.

jtu0:
@thrillberg did some testing in his dev env (THANK YOU):

* Ran `bundle exec rspec`, confirmed everything passes
* Followed the new README directions, confirmed login appears to log in
* Confirmed when logged out, gets redirected to a login page when attempting to visit `/facilities`, `/reports`, or `/file_uploads/new` routes

He ran into one problem:
Running `rake db:seed` throws this error:
```
rake aborted!                                                                                                                     
CsvImporter::InvalidCsvCategoryError: CsvImporter::InvalidCsvCategoryError
```

But the same behavior occurs in `master` as well.

Issue #79, opened 2019-10-27
DevOps: Set up CI/CD


corneliusellen:
- [x] Look into postgres version... do we need to upgrade?
- [x] Decide and obtain new domain name
- [x] Set up continuous integration  to run tests automatically
- [x] Set up continuous deployment (3 applications are: database, background jobs and main application)

anusha5695:
Hi @corneliusellen : Is this open for a PR?

corneliusellen:
@anusha5695 hi! Thanks for your interest. This is actually being working on by the Optoro team, but feel free to pick up a different issue with a 'help wanted' or 'hackoberfest' label.

anusha5695:
@corneliusellen : Sure thanks

PR #91, opened 2019-10-31
Configure Travis to deploy to Heroku on successful master branch builds

yarmiganosca:
Resolves **the CD component** of #79 

### Description

We decided to use Travis' Heroku integration instead of having Heroku pull directly from GitHub b/c we can ensure only successful builds get deployed this way. If Heroku pulls from GitHub whenever `master` is updated, there's the possibility that we could deploy a broken app to Heroku.

### Things That Need To Happen Before Deploying:

* Currently, this is deploying with a Heroku Auth Token tied to my Heroku account, which is not a preferable situation long-term. If Heroku is chosen as the deploy platform for the app, a dedicated Heroku account should be created for it. Once that's done, you should create a Heroku Auth Token for that account and add it as a Travis Project Variable named `HEROKU_DEPLOY_TOKEN`.
* The account the Heroku app is running on is tied to a colleague's credit card. This is also not viable long-term.

### How Has This Been Tested?

It has been deployed from my fork of the abalone repo. Here's the url: https://abalonerescue.herokuapp.com/


yarmiganosca:
I have no clue why the diff looks the way it does.

morhook:
I'm guessing your branch is not updated compared between `forks`. You need to merge branches between forks. Depending on your `remotes` configuration you might have to do something like (standing on your branch `issue-79-travis-heroku-deploy`):

```
git remove -v # this command should show your remotes configuration. At least two makes sense for next commands
git fetch origin # this one could be name different, like main_repo instead
git merge origin/master # here you can get conflicts, resolve them
git push upstream issue-79-travis-heroku-deploy # this one could be called yarmiganosca
```

yarmiganosca:
I plan to investigate why the diff is weird, but I can't until after RubyConf.

jtu0:
@pollygee 

I took a look at this PR, and I think if we plan to release `abalone` for production usage on heroku for some period of time, it makes sense to go ahead and merge this, and stay aware that we'll want to change this as part of the work to change the stack.

If we plan to change platforms _before_ releasing for production usage, it would probably be best to close without merging.

What's the call @pollygee? Should we plan to release to production on heroku (and plan to move the work in this PR later), or should we plan to never release to production on heroku (and close this PR without merging)?

pollygee:
Thanks @jtu0. We should merge this.  We are not ready to move platforms yet.

jtu0:
@corneliusellen I'm going to merge this now - what's the process for deploying this? can you make sure this looks ok post-merge?

PR #92, opened 2019-11-01
[84] Backend/Frontend: Better Real-time CSV Upload Errors for Incorrect Headers [POC]

### LINKED ISSUE #84: Backend/Frontend: Better CSV Upload Errors for Incorrect Headers

rafaltrojanowski:
Resolves #84 

### Description

I used *ActionCable* because it’s free, built-in in Rails, and relatively easy to use.

There are a few drawbacks though:
1) It takes 3-5 seconds to see a notification, it may be a little confusing for users. [Solved]
2) To make integration tests green, there is a need to run `redis-server` before. [Solved]

Things to do:
- [x] Broadcast only to `current_user`
- [x] ~~Add spinner. It would require to broadcast a success message as well. Otherwise in the case of success spinner would never disappear.~~
- [x] Make tests pass without Redis, try to stub ActionCable in the feature tests.
- [x] Add delete button within notification box, need to add Javascript as Bulma does not provide this functionality.

### Type of change

* Documentation update

### How Has This Been Tested?

Manual test, plus `rspec spec`.

Note: It will work only with Tagged/UntaggedAnimalAssessment. For SpawningSuccess more work is needed (it would be better to do it in a separate PR).

There are 3 cases:
1) Processed with success: "Done!" - notification in the green box
2) Invalid headers: "Errors!" Invalid headers: (list of invalid headers)" - in a yellow box
3) File already processed: "Done!" Already processed a file with the same name. Data not imported!" in a green box (screenshot is not up to date)

During processing, there is "Processing..." notification in a blue box.

### Screenshots

processing
success-all-good
invalid-headers
success-already-processed


app/assets/javascripts/channels/notifications.js
```diff
@@ -0,0 +1,16 @@
...
+  },
+
+  disconnected: function() {
+    // Called when the subscription has been terminated by the server
+    // console.log("Disconnected to notifications")
+  },
+
+  received: function(data) {
+    // Called when there's incoming data on the websocket for this channel
+    document.getElementById("notification-message").innerHTML = data.html
```
baweaver: If you send JSON instead it will be easier to test and build up templating on the frontend.

app/models/concerns/import_job.rb
```diff
@@ -35,7 +35,19 @@ def validate_headers(filename)
...
+    unless success
+      ActionCable.server.broadcast "import_job", { html:
+        "<div class='notification is-warning'>
+          <button class='delete'></button>
+          <strong>Invalid headers</strong>
+          <p>Valid headers: #{valid_headers.join(', ')}</p>
+          <p>Current headers: #{headers.join(', ')}</p>
+        </div>"
+      }
+    end
```
baweaver: If you were to express this as JSON, perhaps something like:

```ruby
{
  invalid_headers: headers - valid_headers
  valid_headers: valid_headers
}
```

app/models/concerns/import_job.rb
```diff
@@ -35,7 +35,19 @@ def validate_headers(filename)
...
-    valid_headers == headers
+
+    success = valid_headers == headers
+    unless success
+      ActionCable.server.broadcast "import_job", { html:
+        "<div class='notification is-warning'>
+          <button class='delete'></button>
+          <strong>Invalid headers</strong>
+          <p>Valid headers: #{valid_headers.join(', ')}</p>
+          <p>Current headers: #{headers.join(', ')}</p>
```
baweaver: If you were to go with the JSON route (or really even if you didn't for testing reasons) it may be easier to work with this HTML:


```html
<div class='notification is-warning'>
  <button class='delete'></button>
  
  <div class="notification-error-title">  
    <strong>Error:</strong> <span class="notification-error-type">Invalid Headers</span>
  </div>

  <div class="notification-error-content">
  </div>
</div>
```

...which would make the associated jQuery bind to replacing the content of that something like this:

```javascript
// These two will already make testing easier by being able to pull sections by a specific marker.
const $notificationErrorType = $('.notification-error-type');
const $notificationErrorContent = $('notification-error-content');

const wrap = tagName => contents => `<${tagName}>${Array.from(contents).join('')}</${tagName}>`
const ul = wrap('ul');
const li = wrap('li');

const list = items => ul(items.map(li));

App.notifications = App.cable.subscriptions.create("NotificationsChannel", {
  received ({ invalid_headers, valid_headers }) {
    $notificationErrorContent.html(list(data.invalid_headers));
  }
});
```

Give or take, something like that.

baweaver: That'd also allow you to specify the error type in the JSON response making that a bit more of a universal component that could be used for any type of async error later.

baweaver:
To prevent confusion I would highly suggest using a spinner or other pending action to let users know something is happening or processing.

In Bootstrap that would be something around  the lines of:

```html
<div class="spinner-border" role="status">
  <span class="sr-only">Loading...</span>
</div>
```

https://getbootstrap.com/docs/4.4/components/spinners/

rafaltrojanowski:
@baweaver I appreciate your feedback. I've added a few improvements to the code as you suggested. I went with "Processing" -> "Done" approach instead of a spinner, so the notification content is transmitted from backend too, IMHO it should be okay.

I hope I don't overengineer the backend. I wanted to separate the logic from the import_job and allow notifications to be extended in the future for different contexts. If someone has any suggestions for a better approach, naming, etc let me know.

Thanks again for your suggestions!

app/models/concerns/import_job.rb
rafaltrojanowski: Good idea!

corneliusellen:
@rafaltrojanowski Thank you for your contribution. Your PR actually sparked a discussion around our policy for adding dependencies and other components. We have decided to close this PR for several reasons:
1. Because of the limitations of our stakeholder, it would be best to avoid adding additional dependencies like Redis that may increase operational hosting costs in the future. This was not in our documentation, so thank you for exposing this gap! You can see we've just added to our ReadMe here: https://github.com/rubyforgood/abalone/pull/101
2. Additionally, because of the relatively small file sizes of the lab's CSVs, real-time updates using ActionCable may be over-engineering for this issue. This was also not clear in the original issue. We do appreciate your thoughtfulness around user experience and are open to other solutions that you come up with.

Thanks again for your hard work on this PR and we hope you continue to contribute to the abalone project!

rafaltrojanowski:
@corneliusellen Sure, I would be happy to contribute to the project as long as I can find a time and something that I could help with.

Btw, I'm just seeing that `heroku-redis:hobby-dev` is free, but indeed, there would be a need for something more powerful.

Thanks for your feedback! 🦐

PR #94, opened 2020-01-02
Fix seeds and move invalid csv files to spec dir

corneliusellen:
I broke the seeds when I added new CSV files for categories that we don't yet have an importer for.


Gemfile
```diff
@@ -27,6 +27,8 @@ gem 'devise', '~> 4.7'
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', '>= 1.1.0', require: false
 
+gem 'pry'
```
jtu0: Do you want to keep this in just development/test instead of also in production?

corneliusellen: Good call will move

PR #101, opened 2020-01-02
add architectural constraints and other considerations sections

corneliusellen:
### Description

I added Architectural Constraints and Other Considerations sections based off the discussion we had in our pairing session today.

If you had another example you've seen in the wild of overcomplicating code or not keeping it beginner-friendly, maybe we should add that instead to the ```Other Considerations```section.

### Type of change

* Documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,18 +63,24 @@ annotate
 ```
 Check out their Github page for more running options.
 
+### Architectural Constraints
+Cost is a limitation of this app. We are hosting on a free Heroku instance and need to keep it this way. Do not add additional components that could potentially add costs to hosting (e.g. databases, frameworks, etc). Please talk to us if you have questions.
+
+### Other Considerations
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,18 +63,24 @@ annotate
 ```
 Check out their Github page for more running options.
 
+### Architectural Constraints
+Cost is a limitation of this app. We are hosting on a free Heroku instance and need to keep it this way. Do not add additional components that could potentially add costs to hosting (e.g. databases, frameworks, etc). Please talk to us if you have questions.
```
robbkidd: What do you think of a little softening of the wording?

> In submitting features or bug fixes, please do not add new infrastructure components—e.g. databases, message queues, external caches—that might increase operational hosting costs. Come talk to us if you have questions!

Might add information on where to talk to you (link to Slack? open an issue here?).

corneliusellen: I like it - will update. Thanks for the feedback @robbkidd

PR #103, opened 2020-01-03
Simplify CSV import

robbkidd:
Because the `db/seeds` and the `ImportJob` jobs were both importing CSVs, it made sense to consolidate the business logic of importing CSV data in one place. The `CsvImporter` seems a good place.

* Made `CsvImporter` responsible for doing the actual import

This makes the background jobs a lot simpler by deferring to the importer to do the actual data processing.

`CsvImporter` refactored from a module to a class so that an individual import action is represented by an instance of an importer. A class method `import()` remains that maintains the previous method signature so the existing call sites for CSV import stay the same.

* `CsvImporter::CATEGORIES`

This is a simpler list of the human-readable strings that are the categories of data to import. This is the **One True Place**™ where categories that are importable by CSV are declared. Everywhere that does imports (the import jobs and the seeds) refer to this list to know what's importable.

* Import statistics generation moved to `CsvImporter`

* CSV-to-database-columns translation moved to the models

The `create_from_csv_data()` on the models is now responsible for using a row of CSV data and instantiating a new model ready to be saved to the database. A future improvement to this method could be to perform any data cleansing currently handled by `data_cleanse!()` and save the model to the database.

The `HEADERS` mapping between CSV column names and database column names already lived in the models.  Colocating the mapping and the methods that execute on that mapping should make maintaining them clearer.

### DIFF
--- a/app/controllers/file_uploads_controller.rb
+++ b/app/controllers/file_uploads_controller.rb
@@ -2,12 +2,7 @@ class FileUploadsController < ApplicationController
 
   # The second value for each category entry will be used to determine the job class that processes the data.
   # Ex: Selecting "Spawning Success" in the form will post "SpawningSuccess" and process the data with a SpawningSuccessJob.
-  CATEGORIES = [
-      ['Spawning Success','SpawningSuccess'],
-      ['Tagged Animal Assessment','TaggedAnimalAssessment'],
-      ['Untagged Animal Assessment', 'UntaggedAnimalAssessment'],
-      # ['Wild Collection','WildCollection']
-  ].freeze
+  CATEGORIES = CsvImporter::CATEGORIES.map {|category| [category, category.delete(' ')] }.freeze
 
   def index
     @processed_files = ProcessedFile.all.order(updated_at: :desc).first(20)
--- a/app/controllers/file_uploads_controller.rb
+++ b/app/controllers/file_uploads_controller.rb
@@ -2,14 +2,14 @@ class FileUploadsController < ApplicationController
 
   # The second value for each category entry will be used to determine the job class that processes the data.
   # Ex: Selecting "Spawning Success" in the form will post "SpawningSuccess" and process the data with a SpawningSuccessJob.
-  CATEGORIES = CsvImporter::CATEGORIES.map {|category| [category, category.delete(' ')] }.freeze
+  FILE_UPLOAD_CATEGORIES = CsvImporter::CATEGORIES.map {|category| [category, category.delete(' ')] }.freeze
 
   def index
     @processed_files = ProcessedFile.all.order(updated_at: :desc).first(20)
   end
 
   def new
-    @categories = [['Select One', '']] + CATEGORIES
+    @categories = [['Select One', '']] + FILE_UPLOAD_CATEGORIES
   end
 
   def show_processing_csv_errors
@@ -44,9 +44,8 @@ class FileUploadsController < ApplicationController
   end
 
   def upload
-
     @category = params[:category]
-    if CATEGORIES.map{|label,value| value}.include?(@category)
+    if FILE_UPLOAD_CATEGORIES.map{|label,value| value}.include?(@category)
       uploaded_io = params[:input_file]
       timestamp = Time.new.strftime('%s_%N')
       if uploaded_io
--- a/app/jobs/concerns/import_job.rb
+++ b/app/jobs/concerns/import_job.rb
@@ -3,7 +3,7 @@ module ImportJob
   extend ActiveSupport::Concern
 
   included do |base|
-    attr_accessor :processed_file
+    attr_accessor :processed_file, :stats
 
     base.extend ImportJobClassMethods
   end
@@ -41,17 +41,7 @@ module ImportJob
 
   def import_records(filename)
     raise "No input file specified" unless filename
-    IOStreams.each_record(filename) do |record|
-      attrs = preprocess_attribute_values(translate_attribute_names(record))
-      initialized_model = category_model.new(
-          attrs.merge({processed_file_id: @processed_file.id, raw: false})
-      )
-      initialized_model.cleanse_data!
-      unless initialized_model.save
-        log("Error: Row #{stats[:row_count] + 2} is not valid. #{attrs}", :error)
-      end
-      increment_stats(attrs, initialized_model.persisted?)
-    end
+    @stats = CsvImporter.import(filename, category_model.name.underscore.humanize.titleize, @processed_file.id)
     log(stats, :info)
   end
 
@@ -63,42 +53,15 @@ module ImportJob
     @category_model ||= category.constantize
   end
 
-  def stats
-    @stats ||= Hash.new(0)
-    @stats[:shl_case_numbers] = Hash.new(0) unless @stats.key?(:shl_case_numbers)
-    @stats
-  end
-
-  def increment_stats(attrs, persisted)
-    stats[:row_count] += 1
-    if persisted
-      stats[:rows_imported] += 1
-      stats[:shl_case_numbers][attrs['shl_case_number']] += 1 if attrs.key?('shl_case_number')
-    else
-      stats[:rows_not_imported] += 1
-    end
-  end
-
-  # We're receiving date values in a format different than what ActiveRecord expects, we need a mechanism
-  # for applying transformations to values before they get to the model.
-  def preprocess_attribute_values(attrs)
-    attrs
-  end
-
-  def translate_attribute_names(attrs)
-    attrs
-  end
-
   def already_processed?(filename)
     ProcessedFile.where(status: ['Processed','Processed with errors'] ).where(original_filename: original_filename(filename)).count > 0
   end
 
   def initialize_processed_file(filename)
-    @processed_file = ProcessedFile.create(filename: filename,
-                                           original_filename: original_filename(filename),
-                                           category: category,
-                                           status: 'Running',
-                                           job_stats: stats)
+    @processed_file = ProcessedFile.new(filename: filename,
+                                        original_filename: original_filename(filename),
+                                        category: category,
+                                        status: 'Running')
   end
 
   def complete_processed_file!
--- a/app/jobs/concerns/import_job.rb
+++ b/app/jobs/concerns/import_job.rb
@@ -18,12 +18,12 @@ module ImportJob
       if validate_headers(full_path)
         import_records(full_path)
         complete_processed_file!
+        remove_file!(filename) unless stats[:rows_not_imported] > 0
       else
         log("Error: #{filename} does not have valid headers. Data not imported!", :error)
         fail_processed_file("Does not have valid headers. Data not imported!")
       end
     end
-    remove_file!(filename) unless stats[:rows_not_imported] > 0
     @processed_file.save
   end
 
--- a/app/lib/csv_importer.rb
+++ b/app/lib/csv_importer.rb
@@ -1,41 +1,56 @@
-module CsvImporter
-  CSV_CATEGORIES = {
-      SPAWNING_SUCCESS: 'Spawning Success',
-      TAGGED_ANIMAL_ASSESSMENT: 'Tagged Animal Assessment',
-      UNTAGGED_ANIMAL_ASSESSMENT: 'Untagged Animal Assessment'
-  }.freeze
+class CsvImporter
+  attr_reader :stats, :model, :filename, :processed_file_id
 
-  class InvalidCsvCategoryError < StandardError; end;
+  CATEGORIES = [
+      'Spawning Success',
+      'Tagged Animal Assessment',
+      'Untagged Animal Assessment',
+      'Wild Collection',
+  ].freeze
+
+  class InvalidCategoryError < StandardError; end;
 
   def self.import(filename, category_name, processed_file_id)
-    csv_category_model = model_from_category(category_name)
-    IOStreams.each_record(filename) do |record|
-      attrs = translate_attribute_names(record, category_name)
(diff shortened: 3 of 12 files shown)


app/controllers/file_uploads_controller.rb
```diff
@@ -2,12 +2,7 @@ class FileUploadsController < ApplicationController
 
   # The second value for each category entry will be used to determine the job class that processes the data.
   # Ex: Selecting "Spawning Success" in the form will post "SpawningSuccess" and process the data with a SpawningSuccessJob.
-  CATEGORIES = [
-      ['Spawning Success','SpawningSuccess'],
-      ['Tagged Animal Assessment','TaggedAnimalAssessment'],
-      ['Untagged Animal Assessment', 'UntaggedAnimalAssessment'],
-      # ['Wild Collection','WildCollection']
-  ].freeze
+  CATEGORIES = CsvImporter::CATEGORIES.map {|category| [category, category.delete(' ')] }.freeze
```
robbkidd: Part of the effort to consolidate the list of what categories are importable in one place and referencing that single declaration.

app/jobs/concerns/import_job.rb
```diff
@@ -41,17 +41,7 @@ def validate_headers(filename)
...
-      initialized_model = category_model.new(
-          attrs.merge({processed_file_id: @processed_file.id, raw: false})
-      )
-      initialized_model.cleanse_data!
-      unless initialized_model.save
-        log("Error: Row #{stats[:row_count] + 2} is not valid. #{attrs}", :error)
-      end
-      increment_stats(attrs, initialized_model.persisted?)
-    end
+    @stats = CsvImporter.import(filename, category_model.name.underscore.humanize.titleize, @processed_file.id)
```
robbkidd: Made CsvImporter responsible for doing the actual import. This makes the background jobs a lot dumber by deferring to the importer to do the actual process.

app/jobs/concerns/import_job.rb
```diff
@@ -63,42 +53,15 @@ def category_model
...
-
-  def increment_stats(attrs, persisted)
-    stats[:row_count] += 1
-    if persisted
-      stats[:rows_imported] += 1
-      stats[:shl_case_numbers][attrs['shl_case_number']] += 1 if attrs.key?('shl_case_number')
-    else
-      stats[:rows_not_imported] += 1
-    end
-  end
```
robbkidd: Import stats logic moved to the CsvImporter class.

app/jobs/concerns/import_job.rb
```diff
@@ -63,42 +53,15 @@ def category_model
...
-
-  # We're receiving date values in a format different than what ActiveRecord expects, we need a mechanism
-  # for applying transformations to values before they get to the model.
-  def preprocess_attribute_values(attrs)
-    attrs
-  end
-
-  def translate_attribute_names(attrs)
-    attrs
-  end
```
robbkidd: Translating and pre-processing logic moved to a new `create_from_csv_data` method on the importable models.

app/jobs/concerns/import_job.rb
```diff
@@ -63,42 +53,15 @@ def category_model
...
   def initialize_processed_file(filename)
-    @processed_file = ProcessedFile.create(filename: filename,
-                                           original_filename: original_filename(filename),
-                                           category: category,
-                                           status: 'Running',
-                                           job_stats: stats)
+    @processed_file = ProcessedFile.new(filename: filename,
+                                        original_filename: original_filename(filename),
+                                        category: category,
+                                        status: 'Running')
```
robbkidd: Only instantiate the `ProcessedFile` when initializing. The save occurs in the complete/fail_processed_file! methods after stats from the import are collected.

app/lib/csv_importer.rb
```diff
@@ -1,41 +1,56 @@
...
+class CsvImporter
+  attr_reader :stats, :model, :filename, :processed_file_id
 
-  class InvalidCsvCategoryError < StandardError; end;
+  CATEGORIES = [
+      'Spawning Success',
+      'Tagged Animal Assessment',
+      'Untagged Animal Assessment',
+      'Wild Collection',
+  ].freeze
```
robbkidd: This is the One True Place where categories that are importable by CSV are declared. Everywhere that does imports (the import jobs and the seeds) refer to this list to know what's importable.

app/lib/csv_importer.rb
```diff
@@ -1,41 +1,56 @@
...
+
+  class InvalidCategoryError < StandardError; end;
```
robbkidd: This was just a fiddly change. I removed `Csv` from the error class name because it is already under the `CsvImporter` namespace context.

app/lib/csv_importer.rb
```diff
@@ -1,41 +1,56 @@
...
 
   def self.import(filename, category_name, processed_file_id)
-    csv_category_model = model_from_category(category_name)
-    IOStreams.each_record(filename) do |record|
-      attrs = translate_attribute_names(record, category_name)
-      attrs[:processed_file_id] = processed_file_id
-      csv_row_record = csv_category_model.new(attrs.merge({raw: false}))
-      csv_row_record.cleanse_data! if csv_row_record.respond_to?(:cleanse_data!)
-      csv_row_record.save
+    importer = new(filename, category_name, processed_file_id)
```
robbkidd: I like instances of things, so I converted the previous module method to a class method that creates an instance of an importer to perform the work. The "API" of using the `CsvImporter.import()` method still feels the same in the places it is used.

app/lib/csv_importer.rb
```diff
@@ -1,41 +1,56 @@
...
-    case category_name
-    when CSV_CATEGORIES[:SPAWNING_SUCCESS]
-      SpawningSuccess
-    when CSV_CATEGORIES[:TAGGED_ANIMAL_ASSESSMENT]
-      TaggedAnimalAssessment
-    when CSV_CATEGORIES[:UNTAGGED_ANIMAL_ASSESSMENT]
-      UntaggedAnimalAssessment
+  def model_from_category(category_name)
+    if CATEGORIES.include?(category_name)
+      category_name.delete(' ').constantize
```
robbkidd: Reduces duplication. As new models are added for import, they don't have to be entered in as many places.

app/lib/csv_importer.rb
```diff
@@ -1,41 +1,56 @@
...
-    case category_name
-    when CSV_CATEGORIES[:SPAWNING_SUCCESS]
-      attrs['nbr_of_eggs_spawned'] = attrs.delete('number_of_eggs_spawned_if_female')
+  def increment_stats(model)
+    stats[:row_count] += 1
+    if model.persisted?
+      stats[:rows_imported] += 1
+      stats[:shl_case_numbers][model.shl_case_number] += 1 if model.respond_to?(:shl_case_number)
+    else
+      stats[:rows_not_imported] += 1
```
robbkidd: Pretty much a straight lift of the import stats tracking from `ImportJob`.

app/models/spawning_success.rb
```diff
@@ -31,6 +31,12 @@ class SpawningSuccess < ApplicationRecord
 
   validates :shl_case_number, presence: true
 
+  def self.create_from_csv_data(attrs)
+    attrs['nbr_of_eggs_spawned'] = attrs.delete('number_of_eggs_spawned_if_female')
+    attrs['shl_case_number']     = attrs.delete('shl_number')
+    new(attrs)
```
robbkidd: Part of the move of the CSV-to-model conversion from the jobs to the models.

Note that if you expand the collapsed content of the class above this diff, you'll see the HEADERS mapping between CSV column names and database column names. Colocating the mapping and the methods that execute on that mapping should make maintaining them clearer.

app/models/tagged_animal_assessment.rb
```diff
@@ -48,6 +48,10 @@ class TaggedAnimalAssessment < ApplicationRecord
   validates :measurement_date, :shl_case_number, :spawning_date, :tag, :length, presence: true
   validates :length, numericality: true
 
+  def self.create_from_csv_data(attrs)
+    new(attrs)
```
robbkidd: Part of the move of the CSV-to-model conversion from the jobs to the models.

app/models/untagged_animal_assessment.rb
```diff
@@ -48,12 +48,20 @@ class UntaggedAnimalAssessment < ApplicationRecord
   )
   validates :length, numericality: true
 
+  def self.create_from_csv_data(attrs)
+    new(attrs)
```
robbkidd: Part of the move of the CSV-to-model conversion from the jobs to the models.

app/models/untagged_animal_assessment.rb
```diff
@@ -48,12 +48,20 @@ class UntaggedAnimalAssessment < ApplicationRecord
...
+  end
+
   # Cohort is translated to shl_case_number to compute stats.
   # Here we need to transfer it back to be able to store value in the DB.
   def shl_case_number=(value)
     self.cohort = value
   end
 
+  def shl_case_number
+    self.cohort
```
robbkidd: With `cohort` as the column in the database _and_ in the CSV, this seemed a little cleaner to support referring to an SHL case number when dealing with instances of this model instead of doing translation and retranslation in the import.

db/seeds.rb
```diff
@@ -32,7 +32,7 @@
 #   end
 # end
 
-current_csv_importers = ['Spawning Success','Tagged Animal Assessment','Untagged Animal Assessment']
+current_csv_importers = CsvImporter::CATEGORIES
```
robbkidd: Part of the effort to consolidate the list of what categories are importable in one place and referencing that single declaration.

app/models/wild_collection.rb
```diff
@@ -68,6 +68,17 @@ class WildCollection < ApplicationRecord
...
 
+
+  def self.create_from_csv_data(attrs)
+    attrs['proximity_to_nearest_neighbor']              = attrs.delete('proximity_to_nearest_neighbor_m')
+    attrs['collection_depth']                           = attrs.delete('collection_depth_m')
+    attrs['collection_coordinates']                     = attrs.delete('collection_coodinates')
+    attrs['final_holding_facility_and_date_of_arrival'] = attrs.delete('final_holding_facility__date_of_arrival')
+    attrs['collection_date'] = Date.strptime(attrs['collection_date'], '%m/%d/%y') rescue nil
+    attrs['otc_treatment_completion_date'] = Date.strptime(attrs['otc_treatment_completion_date'], '%m/%d/%y') rescue nil
+    new(attrs)
```
robbkidd: Part of the move of the CSV-to-model conversion from the jobs to the models. This is basically the preprocessing and the translate taken care of.

app/lib/csv_importer.rb
```diff
@@ -1,41 +1,56 @@
...
+    @stats = Hash.new(0)
+    @stats[:shl_case_numbers] = Hash.new(0)
+  end
+
+  def process
+    IOStreams.each_record(filename) do |csv_row|
+      csv_row[:processed_file_id] = processed_file_id
+      csv_row[:raw] = false
+      record = model.create_from_csv_data(csv_row)
+      record.cleanse_data! if record.respond_to?(:cleanse_data!)
```
robbkidd: It's probable that whatever data cleansing is happening here can move to inside the `create_from_csv_data()` methods.

app/controllers/file_uploads_controller.rb
jtu0: I had a little trouble realizing that `CATEGORIES` is a completely different structure from `CsvImporter::CATEGORIES`.
Do you think it would disambiguate if we call this something like
```ruby
FILE_UPLOAD_CATEGORIES = CsvImporter::CATEGORIES.map {|category| [category, category.delete(' ')] }.freeze
```
?

robbkidd: Absolutely! Good point.

robbkidd: Absolutely! Good point.

app/models/wild_collection.rb
```diff
@@ -68,6 +68,17 @@ class WildCollection < ApplicationRecord
   validate :initial_holding_facility_is_valid,
            :final_facility_and_date_of_arrival_is_valid
 
+
+  def self.create_from_csv_data(attrs)
+    attrs['proximity_to_nearest_neighbor']              = attrs.delete('proximity_to_nearest_neighbor_m')
+    attrs['collection_depth']                           = attrs.delete('collection_depth_m')
+    attrs['collection_coordinates']                     = attrs.delete('collection_coodinates')
+    attrs['final_holding_facility_and_date_of_arrival'] = attrs.delete('final_holding_facility__date_of_arrival')
```
rafaltrojanowski: `final_holding_facility__date_of_arrival` Double underscore looks like a typo :)

jtu0:
That's what we thought too! :)

It took four of us putting our heads together to realize that's what the
column looks like after going through the CSV parser.  Come join us in the
Slack! ❤️

app/lib/csv_importer.rb
wesleyeewong: If the data cleansing is minimal, I think it's fine. But if it's some heavy business logic, would the model object be the best place to put it?

jtu0:
@corneliusellen @robbkidd I think this is ready to merge?

app/lib/csv_importer.rb
corneliusellen: I like this too. Using instances feels much more OO to me.

app/models/spawning_success.rb
corneliusellen: I think this is one of the main reasons why before this refactor we were getting confused. This makes way more sense!

Issue #104, opened 2020-01-04
Use the db for temp file storage instead of ActiveStorage


corneliusellen:
ActiveStorage will not work on Heroku for temp file storage for reasons listed here:
https://devcenter.heroku.com/articles/active-storage-on-heroku. Currently file uploading breaks on production for this reason :(

Let's just use a good ole' postgresql table instead called ```TemporaryFile``` to temporarily store the CSV. Workflow should be:
1. When a user uploads a file, add the raw data to this table
2. Kick off the job that processes the file and saves the cleaned data to the db
3. Delete the raw data when the job finishes

PR #106, opened 2020-01-09
Store uploaded csv files in database not on disk

jtu0:
Resolves https://github.com/rubyforgood/abalone/issues/104

### Description
This PR addresses the problem where because a heroku worker runs on a different app, it cannot access the local disk storage of the webapp, where the csv files live.

To solve that, this PR does the following:
* Instead of taking csv uploads and writing it to disk, shove it into a `TemporaryFile` object.
* When the `ImportJob` runs, grab the contents of that `TemporaryFile` object, write it to disk, and continue as before.

Note: this code change doesn't delete the `TemporaryFile` objects or clean what's on disk - let's open that request as a new low-priority issue.

We might want to add some architectural notes to the README about this, but I'd like for either someone else to do that if it's needed for this PR before merge, or to do it as a separate PR.

### Type of change
Behavior should be the same locally, but should go from not-working to working in production

### How To Test

When run locally, app behavior should behave as before.  Verify the upload, job runner, and file processing workflow:

- [ ] Upload a valid CSV
- [ ] Make sure your jobs worker is running
- [ ] After the job is run, check if new entries were created

### DIFF
--- a/spec/factories/temporary_files.rb
+++ b/spec/factories/temporary_files.rb
@@ -1,0 +1,5 @@
+FactoryBot.define do
+  factory :temporary_file do
+    
+  end
+end
--- a/app/controllers/file_uploads_controller.rb
+++ b/app/controllers/file_uploads_controller.rb
@@ -49,12 +49,13 @@ class FileUploadsController < ApplicationController
       uploaded_io = params[:input_file]
       timestamp = Time.new.strftime('%s_%N')
       if uploaded_io
-        @filename = [timestamp, uploaded_io.original_filename].join('_')
-        File.open(Rails.root.join('storage', @filename), 'wb') do |file|
-          file.write(uploaded_io.read)
+        begin
+          temporary_file = TemporaryFile.create(contents: uploaded_io.read)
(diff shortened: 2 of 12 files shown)


spec/factories/temporary_files.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :temporary_file do
+
```
mdworken: @jtu0 Super nit: might be more reflective of real-world data to set contents to empty string by default instead of letting them be nil. 

The biggest impact would be if we ever call string methods on the `contents` for this factory without re-specifying the data - an unreplaced nil would NoMethod but empty string wouldn't - not sure which is preferred.

mdworken: On second thought though, IF we want behavior like the latter, I think we should consider throwing a factory-level error saying to specify it

jtu0: I like this idea - going to merge this now so @corneliusellen will be able to demo with the stakeholder this week, if I had time I would do this though!

Issue #97, opened 2020-01-02
Add MortalityTracking CSV upload job and model validations


corneliusellen:
_Acceptance Criteria:_
- [ ] As a user, I want to be able to upload a CSV of category found in `sample_date_files/mortality_tracking` on the page `/file_uploads/new`.
- [ ] Test should be written for this job similar to `spec/jobs/tagged_animal_assessment_job_spec.rb`
- [ ] Model Validations
- [ ] Unit tests for model validations

**Notes:**
Much of the logic for the file uploading/parsing is already written in the module concern `ImportJob`.

The heaviest lifting will be adding appropriate validations for the ```MortalityTracking``` model. Please refer to the [data dictionary](https://github.com/rubyforgood/abalone/wiki/Abalone-Data-Dictionary) and the notes below for coding these. For an example, please look at: [Example of WildCollection model validations and spec](https://github.com/rubyforgood/abalone/pull/89/commits/9054bf20f821de658a2a028549d4a26ab632c3a9)

**If there is a column  for Facility, let's make these foreign keys to the `Facility.rb` model we already have. There should only be certain facilities (see the `seeds.rb` file and users have the ability to add new ones.) 

`MortalityTracking.rb`
- Required: mortality_date, cohort, shl_case_number, spawning_date, # of morts (some might be unknown see note from lab)
- See data dictionary for specific formats for cohort, shl_case_number, tag
- motality_date and spawning_date should be valid date with month, day, and year
- Note from lab:
_These are the data that are probably going to need the most attention and QA/QC on
our end. One challenge here is that mortalities are currently being tracked on this
datasheet via a single entry for each population/location collected from at a single time
point. When there are multiple tagged animals collected at a single timepoint, all the
tagged animals are all lumped together in a single entry on this data sheet, and their tags are listed in the Notes. One other challenge is that sometimes we don’t know exactly which population a dead animal was from (e.g., it was found dead on the floor or in a sump), but we have some guesses
based on the animal size and/or where it was located. Not sure the best way to code for
that._
--> This brings up one question: for those animals that are lumped together in a single entry (`#_of_morts` > 1), can we create multiple entries in the db for each individual animal, and can we extract each animal's tag from the `Note` column??

edwinthinks:
Hello @corneliusellen, I can give this shot :)

edwinthinks:
Submitted a PR for this - https://github.com/rubyforgood/abalone/pull/114

[...]

PR #90, opened 2019-10-31
[WIP]  Issue 81 - Set up linter

Rarcita:
Resolves #001 

### Type of change

* Bug fix (non-breaking change which fixes an issue)
* New feature (non-breaking change which adds functionality)
* Breaking change (fix or feature that would cause existing functionality to not 
work as expected)
* This change requires a documentation update
* Documentation update

Issue #81, opened 2019-10-27
Set up linter/style rules


corneliusellen:
We want to have a basic linter set up and enforced for new PRs.

yportnoy:
https://github.com/rubocop-hq/rubocop

morhook:
Linking PR: https://github.com/rubyforgood/abalone/pull/90

sarslanoglu:
@corneliusellen Is this issue still not completed? I can work on it if needed

PR #113, opened 2020-01-16
Completely fail on import error

### LINKED ISSUE #102: CSVs with invalid rows should fail on upload
This resulted from a conversation in this PR: https://github.com/rubyforgood/abalone/pull/88

Currently, when a CSV is processed with some invalid rows, we reject the invalid rows and save the valid rows. However, we should not save _any_ rows to the db because this risks the user of duplicating data. Let's reject the entire CSV and tell the user to fix the invalid rows and re-upload the file.

wesleyeewong:
Resolves #102  

### Description

Feature to rollback all created records if we encounter a single failed import when processing a file.

Refactored `import_job.rb` to support such failure, report appropriate fail message, and set appropriate file status.

### Type of change

work as expected)

### How Has This Been Tested?

Wrote spec test for `CsvImporter`, and tested it locally.

### Screenshots
Screenshot 2020-01-16 at 3 29 58 PM

### DIFF
--- a/app/jobs/concerns/import_job.rb
+++ b/app/jobs/concerns/import_job.rb
@@ -4,6 +4,7 @@ module ImportJob
 
   included do |base|
     attr_accessor :processed_file, :stats
+    attr_reader :job_stats
 
     base.extend ImportJobClassMethods
   end
@@ -12,23 +13,30 @@ module ImportJob
     filename = args[0]
     full_path = Rails.root.join('storage', filename).to_s
     initialize_processed_file(filename)
+
     if already_processed?(filename)
       fail_processed_file("Already processed a file with the same name. Data not imported!")
     else
       if validate_headers(full_path)
-        import_records(full_path)
-        complete_processed_file!
-        remove_file!(filename) unless stats[:rows_not_imported] > 0
+        if import_records(full_path)
+          complete_processed_file!
+          remove_file!(filename)
+        else
+          log("Error: #{filename} does not have valid row(s). Data not imported!", :error)
+          fail_processed_file("Does not have valid row(s). Data not imported!")
+        end
       else
-        log("Error: #{filename} does not have valid headers. Data not imported!", :error)
-        fail_processed_file("Does not have valid headers. Data not imported!")
+        log("Error: #{filename} does not have valid header(s). Data not imported!", :error)
+        fail_processed_file("Does not have valid header(s). Data not imported!")
       end
     end
+
     @processed_file.save
   end
 
   def validate_headers(filename)
     raise "No input file specified" unless filename
+
     headers = CSV.parse(File.read(filename, encoding: 'bom|utf-8'), headers: true)
       .headers
       .compact
@@ -41,8 +49,16 @@ module ImportJob
 
   def import_records(filename)
     raise "No input file specified" unless filename
-    @stats = CsvImporter.import(filename, category_model.name.underscore.humanize.titleize, @processed_file.id)
-    log(stats, :info)
+    csv_importer = CsvImporter.new(filename, category_model.name.underscore.humanize.titleize, @processed_file.id)
+    csv_importer.call
+
+    if csv_importer.erred?
+      return false
+    end
+
+    @job_stats = csv_importer.stats
+    log(job_stats, :info)
+    true
   end
 
   def category
@@ -65,12 +81,8 @@ module ImportJob
   end
 
   def complete_processed_file!
-    @processed_file.job_stats = stats
-    if stats[:rows_not_imported] > 0
-      @processed_file.status = 'Processed with errors - only re-upload fixed rows to avoid data duplication'
-    else
-      @processed_file.status = 'Processed'
-    end
+    @processed_file.job_stats = job_stats
+    @processed_file.status = 'Processed'
   end
 
   def fail_processed_file(error)
--- a/app/jobs/concerns/import_job.rb
+++ b/app/jobs/concerns/import_job.rb
@@ -4,7 +4,7 @@ module ImportJob
 
   included do |base|
     attr_accessor :processed_file, :stats
-    attr_reader :job_stats
+    attr_reader :job_stats, :error_details
 
     base.extend ImportJobClassMethods
   end
@@ -63,6 +63,8 @@ module ImportJob
     csv_importer.call
 
     if csv_importer.errored?
+      @error_details = csv_importer.error_details
+      log(error_details, :info)
       return false
     end
 
@@ -96,7 +98,7 @@ module ImportJob
 
   def fail_processed_file(error)
     @processed_file.status = 'Failed'
-    @processed_file.job_stats = {}
+    @processed_file.job_stats = error_details || {}
     @processed_file.job_errors = error
   end
 
--- a/app/jobs/concerns/import_job.rb
+++ b/app/jobs/concerns/import_job.rb
@@ -62,7 +62,7 @@ module ImportJob
     csv_importer = CsvImporter.new(filename, category_model.name.underscore.humanize.titleize, @processed_file.id)
     csv_importer.call
 
-    if csv_importer.erred?
+    if csv_importer.errored?
       return false
     end
 
--- a/app/lib/csv_importer.rb
+++ b/app/lib/csv_importer.rb
@@ -26,15 +26,22 @@ class CsvImporter
   end
 
   def process
-    IOStreams.each_record(filename) do |csv_row|
-      csv_row[:processed_file_id] = processed_file_id
-      csv_row[:raw] = false
-      record = model.create_from_csv_data(csv_row)
-      record.cleanse_data! if record.respond_to?(:cleanse_data!)
-      record.save
-      increment_stats(record)
+    model.transaction do
+      IOStreams.each_record(filename) do |csv_row|
+        csv_row[:processed_file_id] = processed_file_id
+        csv_row[:raw] = false
+        record = model.create_from_csv_data(csv_row)
+        record.cleanse_data! if record.respond_to?(:cleanse_data!)
+        unless record.save
+          @stats = Hash.new(0)
+          raise ActiveRecord::Rollback
+        end
+        increment_stats(record)
+      end
     end
   end
+
+  private
 
   def model_from_category(category_name)
     if CATEGORIES.include?(category_name)
--- a/app/lib/csv_importer.rb
+++ b/app/lib/csv_importer.rb
@@ -39,11 +39,11 @@ class CsvImporter
         record = model.create_from_csv_data(csv_row)
         record.cleanse_data! if record.respond_to?(:cleanse_data!)
 
-        unless record.valid?
-          error_details["row_number_#{row_number}"] = record.errors.details
-        else
+        if record.valid?
           record.save
           increment_stats(record)
+        else
+          error_details["row_number_#{row_number}"] = record.errors.details
         end
 
         row_number += 1
--- a/app/lib/csv_importer.rb
+++ b/app/lib/csv_importer.rb
@@ -1,6 +1,6 @@
 class CsvImporter
-  attr_reader :stats, :model, :filename, :processed_file_id, :erred
-  alias :erred? :erred
+  attr_reader :stats, :model, :filename, :processed_file_id, :errored
+  alias :errored? :errored
 
   CATEGORIES = [
       'Spawning Success',
@@ -17,7 +17,7 @@ class CsvImporter
     @model = model_from_category(category_name)
     @stats = Hash.new(0)
     @stats[:shl_case_numbers] = Hash.new(0)
-    @erred = false
+    @errored = false
   end
 
   def call
@@ -36,7 +36,7 @@ class CsvImporter
 
         unless record.save
           @stats = Hash.new(0)
-          @erred = true 
+          @errored = true 
           raise ActiveRecord::Rollback
         end
 
--- a/app/lib/csv_importer.rb
+++ b/app/lib/csv_importer.rb
@@ -1,6 +1,5 @@
 class CsvImporter
-  attr_reader :stats, :model, :filename, :processed_file_id, :errored
-  alias :errored? :errored
+  attr_reader :stats, :model, :filename, :processed_file_id, :error_details
 
   CATEGORIES = [
       'Spawning Success',
@@ -16,17 +15,23 @@ class CsvImporter
     @processed_file_id = processed_file_id
     @model = model_from_category(category_name)
     @stats = Hash.new(0)
+    @error_details = {}
     @stats[:shl_case_numbers] = Hash.new(0)
-    @errored = false
   end
 
   def call
     process
   end
 
+  def errored?
+    !error_details.empty?
+  end
+
   private
 
   def process
+    row_number = 2 # assuming 1 is headers
+
     model.transaction do
       IOStreams.each_record(filename) do |csv_row|
         csv_row[:processed_file_id] = processed_file_id
@@ -34,13 +39,19 @@ class CsvImporter
         record = model.create_from_csv_data(csv_row)
         record.cleanse_data! if record.respond_to?(:cleanse_data!)
 
-        unless record.save
-          @stats = Hash.new(0)
-          @errored = true 
-          raise ActiveRecord::Rollback
+        unless record.valid?
+          error_details["row_number_#{row_number}"] = record.errors.details
+        else
+          record.save
+          increment_stats(record)
         end
 
-        increment_stats(record)
+        row_number += 1
+      end
+
+      unless error_details.empty?
+        @stats = Hash.new(0)
+        raise ActiveRecord::Rollback
       end
     end
   end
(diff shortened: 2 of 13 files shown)


db/sample_data_files/wild_collection/example_of_wild_collection_data.csv
```diff
@@ -1,14 +1,14 @@
 Tag,Collection_date,General Location,Precise Location,Collection_coodinates,Proximity to nearest neighbor (m),Collection method notes,Foot condition notes,Collection_depth_(m),Length,Weight,Gonad Score,Predicted Sex,Initial holding facility,Final holding facility & date of arrival,OTC treatment completion date
 Green_314,10/21/16,Los Angeles County,redacted,"1,1",>10,spatula,no cuts or abrasians reported,21.6,175,1242,1,M,AoP,BML; 12/13/2016,NA
```
wesleyeewong: actually caught some invalid data :) (I hope)

Can someone confirm

spec/jobs/wild_collection_job_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:filename) { "example_of_wild_collection_data.csv" }
 
   before do
-    Rails.application.load_seed if Facility.valid_codes.empty?
+    Rails.application.load_seed
```
wesleyeewong: I'm not sure if this is the right approach, for some reason, the cache is already loaded with some Facilities, but not all are present, causing the tests to fail

wesleyeewong:
This is ready for review 👍

app/lib/csv_importer.rb
```diff
@@ -26,16 +26,23 @@ def initialize(filename, category_name, processed_file_id)
...
 
   def process
-    IOStreams.each_record(filename) do |csv_row|
-      csv_row[:processed_file_id] = processed_file_id
-      csv_row[:raw] = false
-      record = model.create_from_csv_data(csv_row)
-      record.cleanse_data! if record.respond_to?(:cleanse_data!)
-      record.save
-      increment_stats(record)
+    model.transaction do
```
jtu0: :+1:

spec/jobs/wild_collection_job_spec.rb
jtu0: Does `load_seed` do any kind of dupe suppression? would we likely run into any problems with possibly loading db seeds twice?

app/lib/csv_importer.rb
```diff
@@ -1,5 +1,6 @@
 class CsvImporter
-  attr_reader :stats, :model, :filename, :processed_file_id
+  attr_reader :stats, :model, :filename, :processed_file_id, :erred
+  alias :erred? :erred
```
jtu0: How would you feel about changing this name to `errored`?  I agree that `erred` is more correct, but I wonder if `errored` might be more readable?

db/sample_data_files/wild_collection/example_of_wild_collection_data.csv
jtu0: ^ @corneliusellen @jcavena can either of you weigh in on if `redacted` should be invalid for `Precise Location`?  I don't know the project well enough to know.

app/lib/csv_importer.rb
```diff
@@ -10,29 +11,37 @@ class CsvImporter
 
   class InvalidCategoryError < StandardError; end;
 
-  def self.import(filename, category_name, processed_file_id)
```
jtu0: :+1: Thanks for the refactor here!

spec/jobs/wild_collection_job_spec.rb
wesleyeewong: That I don't know, will have to look into it.

app/lib/csv_importer.rb
wesleyeewong: I'll change it to `errorred`

spec/jobs/wild_collection_job_spec.rb
wesleyeewong: Looking at the doc, it should be ok: https://apidock.com/rails/ActiveRecord/Relation/load

But also tested it out like so:
```
  before do
    Rails.application.load_seed
  end

  before(:each) do
    Rails.application.load_seed
  end
```

And it doesn't dupe the records

wesleyeewong:
I have addressed some of the comments, please let me know if there's anything else I need to address :)

spec/jobs/wild_collection_job_spec.rb
jtu0: Thank you for checking! :heavy_check_mark::heavy_check_mark:

jtu0:
LGTM though I'd feel a lot more comfortable if someone more familiar with the project took a look before merging.  Feel free to merge, or I'll merge late tonight.

wesleyeewong:
> LGTM though I'd feel a lot more comfortable if someone more familiar with the project took a look before merging. Feel free to merge, or I'll merge late tonight.

Besides the question about the `Precise Location` column, I'm fine with merging it. I don't have merge privilege so someone else would have to do it.

db/sample_data_files/wild_collection/example_of_wild_collection_data.csv
corneliusellen: @wesleyeewong ```redacted``` is intentional. The precise lat/long location is confidential data.

wesleyeewong: ok, but would end users also be uploading the lat/long as `redacted`? Or the actual lat/long?

wesleyeewong: If end users will be uploading precise lat/longs then we should validate it, by faking the data here. If it'll always be `redacted`, then there's no reason to store that in the db.

wesleyeewong:
> @wesleyeewong Thank you for your work on this issue.
> 
> I'd like the user to know which specific rows were invalid (this might not have been clear on the original [issue](https://github.com/rubyforgood/abalone/issues/102)). So if there were multiple invalid rows we'd want the user to see all of them on the first failed upload, which is why the #increment_stats method is important to keep.
> 
> Would ActiveRecord rollback would allow us to surface multiple erred rows? If so, I think you could make a few changes to this PR before we merge it in. If not, you might have to switch gears and go with your idea 2 comment on the issue.

Gotcha, Rollback would be called immediately on the first encounter of invalid rows. So most likely option 2 :) Actually it'll be a combination, of 2 and 1.

I'll check for validity using `.valid?` . If valid, save, if not record the row and continue, then at the end, we rollback.

Thanks for letting me know. 👏 👏

wesleyeewong:
@corneliusellen  I've implemented the requested changes, error details will be shown under `stats` like so:

Screenshot 2020-01-22 at 5 21 38 PM

app/lib/csv_importer.rb
```diff
@@ -10,29 +10,49 @@ class CsvImporter
...
+          increment_stats(record)
+        end
+
+        row_number += 1
+      end
+
+      unless error_details.empty?
+        @stats = Hash.new(0)
+        raise ActiveRecord::Rollback
+      end
```
rafaltrojanowski: interesting solution :)

app/lib/csv_importer.rb
```diff
@@ -10,29 +10,49 @@ class CsvImporter
...
+        csv_row[:raw] = false
+        record = model.create_from_csv_data(csv_row)
+        record.cleanse_data! if record.respond_to?(:cleanse_data!)
+
+        unless record.valid?
+          error_details["row_number_#{row_number}"] = record.errors.details
+        else
+          record.save
+          increment_stats(record)
+        end
```
rafaltrojanowski: https://github.com/rubocop-hq/ruby-style-guide#no-else-with-unless

jtu0: Nice catch @rafaltrojanowski!

If you're so inclined, it'd be super helpful if you could help get style catches automated:
https://github.com/rubyforgood/abalone/issues/81 (previous WIP on this here: https://github.com/rubyforgood/abalone/pull/90)

jtu0:
Thanks for the new commit @wesleyeewong!  It looks good to me :+1:

app/lib/csv_importer.rb
wesleyeewong: used `if...else` instead. Still trying to get used to RoR best practices 😄  Thanks for letting me know.

app/jobs/concerns/import_job.rb
```diff
@@ -53,8 +59,18 @@ def validate_headers(full_path)
...
+
+    if csv_importer.errored?
+      @error_details = csv_importer.error_details
+      log(error_details, :info)
+      return false
+    end
+
+    @job_stats = csv_importer.stats
+    log(job_stats, :info)
+    true
```
edwinthinks: Would this be better placed in the `else` block of the code above? I noticed that the code above short circuits and then this would never be reached. I think this is primarily a preference so I think it is fine either way.

spec/lib/csv_importer_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "rails_helper"
+
+RSpec.describe CsvImporter do
```
edwinthinks: Great specs! I dig it!

app/lib/csv_importer.rb
```diff
@@ -10,29 +10,49 @@ class CsvImporter
...
+        csv_row[:raw] = false
+        record = model.create_from_csv_data(csv_row)
+        record.cleanse_data! if record.respond_to?(:cleanse_data!)
+
+        if record.valid?
+          record.save
+          increment_stats(record)
+        else
+          error_details["row_number_#{row_number}"] = record.errors.details
+        end
```
edwinthinks: Thoughts on using a `begin/rescue` /w `save!` approach here and keeping all else the same? I've noticed that sometimes `create_from_csv_data` will raise an error if the DateTime is not parsed correctly. 

For example in `app/models/tagged_animal_assessment.rb:55` if the value of `measurement_date_str` is not valid it will trigger a raised error.

I discovered this while working on https://github.com/rubyforgood/abalone/issues/97

db/schema.rb
```diff
@@ -146,7 +146,7 @@
     t.decimal "growout_trough"
     t.decimal "length"
     t.decimal "mass"
-    t.string "gonad_score"
+    t.decimal "gonad_score"
```
robbkidd: ⚠️ This is a schema change in the PR, currently without a corresponding database migration to change the column type in existing databases.

app/jobs/concerns/import_job.rb
jtu0: I like the way it's currently structured - I feel like putting everything else into an `else` after a guard `if` can make the code harder to read because you have to remember why the code is nested.

app/lib/csv_importer.rb
jtu0: I think that's the direction @wesleyeewong had gone earlier, but @corneliusellen pointed out how it would make it harder for users to understand the problem.

db/schema.rb
jtu0: Oops, good catch @robbkidd!

@wesleyeewong Can you fix this?  I think this PR is ready to merge aside from that.

Also I know it's been a really long-running PR with a ton of requests for change, and you've put a lot of really patient work into it, so if you want any help getting the last little bit done, please ask :heart:

spec/jobs/wild_collection_job_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:filename) { "example_of_wild_collection_data.csv" }
 
   before do
```
mdworken: Sorry for the nit, but because before(:each) rolls back and before(:all) doesn't, it would be really helpful to make it clearer which behavior we're using here. From a quick SO search I *think* `before` defaults to `before(:each)`, but I'm still not 100% sure of any edge cases etc. If we're now always expanding the use of the hook, it would be nice to know for sure how it's behaving without the next person who sees this having to google-fu.

```suggestion
   before(:each) do
```

wesleyeewong:
Thanks guys, I'll look into the comments, and make the necessary changes. Don't usually work on the weekends, apologies for the late replies.

spec/jobs/wild_collection_job_spec.rb
wesleyeewong: You're right it defaults to `before(:each)`.  And I think that's unnecessary, will make it more clear and change it to `before(:all)`.

Thanks!

spec/jobs/wild_collection_job_spec.rb
wesleyeewong: I was mistaking and thought `before` defaults to `default(:all)`. 

Tested it like so:
```
before(:all) do
...
end

before(:each) do
...
end
```

Result remains the same. 👍

mdworken: As a note, :all does NOT automatically rollback any changes in the DB. I think that's fine here, though. Just including this comment for completeness 🎈

app/lib/csv_importer.rb
wesleyeewong: @edwinthinks  ah I see what you mean, it fails before hitting the `valid?` check, on the record creation line `model.create_from_csv_data(csv_row)`

Error reproducible by changing `Measurement_date` in `spec/support/csv/Tagged_assessment_invalid_values.csv` to a non date format string.

I don't think `begin/rescue` would solve the issue tho. 🤔 And I'm not sure what the best solution would be.

spec/jobs/wild_collection_job_spec.rb
wesleyeewong: Oh, good to know. I haven't ran into any issues yet it not being rolled back automatically. Will keep that in mind.

app/models/tagged_animal_assessment.rb
```diff
@@ -45,21 +45,28 @@ class TaggedAnimalAssessment < ApplicationRecord
...
-    write_attribute(:spawning_date, DateTime.strptime(spawning_date_str, '%m/%d/%y'))
+  DATE_ATTRIBUTES.each do |name|
+    define_method "#{name}=" do |argument|
+      return unless argument
+      begin
+        write_attribute(name.to_sym, DateTime.strptime(argument, "%m/%d/%y"))
+      rescue ArgumentError
+        errors.add(name.to_sym, :invalid, message: "Invalid date format: #{argument}")
+      end
+    end
```
wesleyeewong: @edwinthinks I think this should address your concern regarding Date parsing errors.

I'm not sure if there's a better way to do this? Or to dry it up some more? Because I have codes repeating for tagged and untagged models.

I tried extracting the metaprogramming part out into a module like so:

```
module DateAssignment
  extend ActiveSupport::Concern

  DATE_ATTRIBUTES.each do |name|
    ...
   end
end
```

but can't seem to get the constant's scope figured out.

spec/support/csv/Tagged_assessment_invalid_values.csv
```diff
@@ -0,0 +1,5 @@
+Measurement_date,SHL_case_number,Spawning_date,Tag,From_Growout_Rack,From_Growout_Column,From_Growout_Trough,To_Growout_Rack,To_Growout_Column,To_Growout_Trough,Length,Gonad_Score,Predicted_Sex,Notes
+hello,SF16-9A,3/2/16,Lav_028,3,A,1,1,what's,the,good,word?,M,
```
wesleyeewong: @edwinthinks  invalid values that you can test out.

wesleyeewong:
This is ready for another round of review 💣

spec/support/csv/Tagged_assessment_invalid_values.csv
```diff
@@ -1,5 +1,5 @@
 Measurement_date,SHL_case_number,Spawning_date,Tag,From_Growout_Rack,From_Growout_Column,From_Growout_Trough,To_Growout_Rack,To_Growout_Column,To_Growout_Trough,Length,Gonad_Score,Predicted_Sex,Notes
 12/17/18,SF16-9A,3/2/16,Lav_028,3,A,1,1,what's,the,good,word?,M,
 12/17/18,SF16-9A,3/2/16,Lav_057,3,A,to hell with georgia,1,A,2,49,1,M,
-12/17/18,SF16-9A,3/2/16,Lav_029,3,A,1,1,A,2,53.5,1,F,
-12/17/18,SF16-9A,3/2/16,Lav_078,3,A,1,1,A,2,50,1,F,
+12/17/18,SF16-9A,3/2/16,Lav_029,3,A,1,1,how,'bout,them,dawgs?,F,
+12/17/18,SF16-9A,3/2/16,Lav_078,3,A,1,1,A,piss,on,them!,F,
```
corneliusellen: Is this entire test suite a football reference? Let the record show I don't condone any sportsball teams bashing.
But if it's referring to just dogs I'll let it slide.
🐱 > 🐶

corneliusellen: Is this entire test suite a football reference? Let the record show I don't condone any sportsball teams bashing.
But if it's referring to just dogs I'll let it slide.
🐱> 🐶

db/sample_data_files/wild_collection/example_of_wild_collection_data.csv
corneliusellen: TBH I'm not 100%, but I think it will be uploaded as 'redacted'. I will check with the stakeholder. Thanks for bringing this to my attention.

PR #114, opened 2020-01-25
[Issue 97] Add MortalityTracking and job

edwinthinks:
Hello there, this is a PR to address the issue - [here](https://github.com/rubyforgood/abalone/issues/97)

Changes
- Added MortalityTracking and MortalityTrackingJob
- Added Mortality Tracking to the File Upload dropdown options 
- Added `mortality_date=` and `spawning_date=` (modeled after `measurement_date=` in app/models/tagged_animal_assessment.rb). But added a `rescue` to handle cases where the date is not formatted.

## Need Some Help On These Points
- I noticed that `mortality_date` is required according to the issue description but the column `Mortality_date` in the sample contained 'unknown' and 'unkown'. How should we handle this?
- `number_morts` (# Morts) is not always a numerical value. There are 'TBD' and '9+'. How shall we handle this?

### Update
For now, I've removed the rows from the sample data that causes a spec to fail.


app/models/mortality_tracking.rb
```diff
@@ -43,4 +43,38 @@ class MortalityTracking < ApplicationRecord
...
+  def mortality_date=(mortality_date_str)
+    return unless mortality_date_str
+    write_attribute(:mortality_date, DateTime.strptime(mortality_date_str, '%m/%d/%y'))
+  rescue
+    write_attribute(:mortality_date, nil)
+  end
+
+  def spawning_date=(spawning_date_str)
+    return unless spawning_date_str
+    write_attribute(:spawning_date, DateTime.strptime(spawning_date_str, '%m/%d/%y'))
```
mdworken: @edwinthinks what would your thoughts be on using `DateTime#parse` instead here to provide more flexibility in formatting?

app/models/mortality_tracking.rb
```diff
@@ -43,4 +43,38 @@ class MortalityTracking < ApplicationRecord
...
+  rescue
+    write_attribute(:spawning_date, nil)
```
mdworken: I'm not sure how I feel about this rescue here - if DateTime fails to parse it's input, then we'll fail silently here. Should we maybe surface this error to the user somehow?

app/models/mortality_tracking.rb
edwinthinks: I don't mind using `parse`. However, I wonder if we gain some precision by specifying the format of the time? Fine either way :)

app/models/mortality_tracking.rb
edwinthinks: Good point. To be honest, I don't think this conversion of string to date should be in this method. I sense this is going beyond the responsibilities of the model and taking the role of converting a string to date behind the scenes. Instead, I'd put it in the `create_from_csv_data`  (even better in service object IMO) method and let it raise an error if there is an issue.

I'm going to make some modifications and see if it would fit. Though it might not work well for the changes in https://github.com/rubyforgood/abalone/pull/113 since it does not rescue.

app/models/mortality_tracking.rb
edwinthinks: Ahh super interesting. I didn't realize this but `DateTime.parse` has an issue with the date '4/23/01'. Aka
```
DateTime.parse('4/23/01')
```
will raise an error. Going to continue using the original method.

edwinthinks:
@mdworken made some changes since your last comment. Let me know what you think :)

db/sample_data_files/mortality_tracking/example_of_mortality_tracking_data.csv
```diff
@@ -1,2234 +1,1966 @@
-Mortality_date,Cohort,SHL number,Spawning_date,Shell_box,Shell_container,Animal_location,# Morts,Approximation?,Processed by SHL?,Initials,Tag(s),Comments,,,,,,,,,,,,,,,
-9/11/16,2001 Cohort ,SF08-26,4/23/01,SHL,SHL,Broodstock Tank 3,1,,Y,EH,Green_323,"SF08-26-19. Shrunken, Blythe processed",,,,,,,,,,,,,,,
```
mdworken: @edwinthinks just to confirm, all you changed here was the removal of trailing `,`?

mdworken:
@corneliusellen is it okay with you if we spin off the remaining concerns for this ticket into another one? If so, I'll merge this once @edwinthinks pulls the WIP title off.

app/models/mortality_tracking.rb
wesleyeewong: @edwinthinks my PR got merged, and it contains my solution to this issue you're having. I agree that these code should be moved out of the `model` object.

Where I work, we use an `interactor` object for these types of stuff. It goes: `controller` -> `interactor` -> `model`

In `interactor` is where we will do all the validation, and data manipulation/business rules if needed.

But for now, there's not a lot of business logic going on, and they're pretty basic, maybe in the future we can think of refactoring it out into another object.

app/models/mortality_tracking.rb
```diff
@@ -43,4 +43,34 @@ class MortalityTracking < ApplicationRecord
...
+    attrs['number_morts']                     = attrs.delete('_morts')
+
+    # Attempt to convert provided csv attributes
+    # to the proper format. Dates are provided as
+    # strings and must be converted to a datetime.
+    #
+    # Note - DateTime.strptime will raise a ArgumentError
+    # if there is a issue converting.
+    attrs['spawning_date'] = DateTime.strptime(attrs.delete('spawning_date'), '%m/%d/%y')
+    attrs['mortality_date'] = DateTime.strptime(attrs.delete('mortality_date'), '%m/%d/%y')
```
wesleyeewong: https://github.com/wesleyeewong/abalone/blob/e3b30aa4374657e79263968a84170de2eaa498e0/app/models/untagged_animal_assessment.rb#L71-L80

If you add this to your code, the errors will bubble up to user level, where they can discern what's wrong with the csv.

corneliusellen:
@edwinthinks Thanks so much for you hard work on this and implementing feedback. Everything looks good to me, but let's implement @wesleyeewong 's comment about surfacing the date parsing errors since I just merged that PR yesterday. Once that's done I'll merge this.

@mdworken Thank you for reviewing and bringing up the point about surfacing the data parsing errors to the user.

PR #115, opened 2020-02-02
[Issue 112] Create job called by daily cron that deletes processed TemporaryFiles

### LINKED ISSUE #112: Create job called by daily cron that deletes processed TemporaryFiles
When a user uploads a csv, it's loaded into the database to the `temporary_files` table.
Write a job that
- Gets all TemporaryFiles created > 1.week.ago
- Filters out any that don't have a corresponding ProcessedFile or whose ProcessedFile isn't successful (maybe log those so we know about it and can think about how we want to handle them)
- Delete all TemporaryFiles that are older than 1 week old and have been successfully processed

colinsoleim:
This PR:
- Resolves #112
- Creates a rake task to delete or warn about stale TemporaryFiles
- Adds a has_one/belongs_to relationship between temporary and processed files
- Adds rspec tests for the new rake task
- Creates a shared context in rspec for testing future rake tasks

Type of change:
New feature (non-breaking change which adds functionality)

Discussion Points:
- On the data side, I noticed there was no relationship between ProcessedFiles and TemporaryFiles in their respective models. I created a has_one/belongs_to relationship between the two for easier lookup in the rake task. Please let me know if this should be one-to-many instead, or if there was a reason to not have a relationship specified in the model.
- From a feature perspective, I didn't see a way for a user to delete a ProcessedFile that failed to upload, and because the rake task currently sends out warnings about these, there may be a large build up of warnings over time. If this is an issue, I could change the rake task to log temporary files that are completely missing a processed file, or I could add a button to let users delete their failed file uploads.

### DIFF
--- a/app/models/temporary_file.rb
+++ b/app/models/temporary_file.rb
@@ -1,2 +1,5 @@
+# frozen_string_literal: true
+
 class TemporaryFile < ApplicationRecord
+  has_one :processed_file, inverse_of: :temporary_file
 end
--- a/spec/lib/tasks/scheduler_rake_spec.rb
+++ b/spec/lib/tasks/scheduler_rake_spec.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'scheduler:delete_stale_temporary_files' do
+  include_context 'rake'
+
+  let(:temp_file) do
+    create(:temporary_file, contents: '', created_at: 8.days.ago)
+  end
+
+  it 'deletes a stale temporary file that is successfully processed' do
+    create :processed_file, status: 'Processed', temporary_file_id: temp_file.id
+    expect(TemporaryFile.count).to eq 1
+    subject.invoke
+    expect(TemporaryFile.count).to eq 0
+  end
+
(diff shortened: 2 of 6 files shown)


app/models/temporary_file.rb
```diff
@@ -1,2 +1,5 @@
+# frozen_string_literal: true
+
 class TemporaryFile < ApplicationRecord
+  has_one :processed_file
```
mdworken: I'm not sure if it's still necessary for 1-1 relationships in rails 5, but just to be safe, can we add a `inverse_of` here?

https://guides.rubyonrails.org/association_basics.html#bi-directional-associations

spec/lib/tasks/scheduler_rake_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
```
mdworken: A+ specs here!

colinsoleim:
@mdworken Thanks for the review, I added the inverse_of to the processed/temporary file models. About the job itself, Ellen mentioned on discord that she was planning to use Heroku Scheduler to run the rake task.

Issue #69, opened 2019-10-05
Backend/Frontend - Create Multiple Select List and Date Picker for Histogram Parameters


corneliusellen:
There should be an area where the user can select parameters in the upper right hand side of the histogram chart at `/reports` under the Growth tab. The user should have the ability to select a measuring event (i.e. a certain population on a certain date) or a group of measuring events (different populations near the same date).

_Acceptance Criteria:_

- [ ] Backend - returns a list of cohorts 
- [ ] Frontend - Multiple select list of cohorts
- [ ] Frontend - Date picker

We use [Highcharts](https://www.highcharts.com/), a JavaScript library for rendering all graphs and charts.

colinsoleim:
I'll work on this issue next!

Issue #68, opened 2019-10-05
Frontend - Render Histogram with Correct Data


corneliusellen:
BLOCKED by Issue #58 : We need the correct data on the backend first.

We need to make sure the Length Histogram is rendering correctly at `/reports` under the Growth tab. The current report has incorrect data AND incorrect formatting.

_Acceptance Criteria:_

- [ ] Y-axis: Count
- [ ] X-axis: Length in centimeters.
- [ ] Bins should be in 1 cm increments, so 0-0.99 cm, 1-1.99 cm, etc. from 0 to 30cm.
- [ ] Title: Size Distribution
- [ ] Tooltip: When hovering over a bar, user should see the count for that bin increment.

We use [Highcharts](https://www.highcharts.com/), a JavaScript library for rendering all graphs and charts.

PR #117, opened 2020-02-03
68 render histgram correct format

### LINKED ISSUE #66: Backend - Create UntaggedAnimalAssessmentJob
Most of the code for this CSV import is already written in `ImportJob`. Import this module into an `UntaggedAnimalAssessmentJob` and write any custom methods; it should look similar to the  `TaggedAnimalAssessmentJob`.

corneliusellen:
Resolves #66 and #68  

### Description

See issue descriptions for examples of how to arrive at PopulationCountEstimate and the array of lengths for the histogram.

### Type of change

### How Has This Been Tested?

I have only tested this in the UI with hardcoded parameters in the ReportsController. The next step to make this complete is actually passing in these parameters in Issue #69 .
I have not written model tests yet.


app/models/mortality_tracking.rb
```diff
@@ -53,24 +53,11 @@ def self.create_from_csv_data(attrs)
...
     # if there is a issue converting.
-    attrs['spawning_date'] = DateTime.strptime(attrs.delete('spawning_date'), '%m/%d/%y')
-    attrs['mortality_date'] = DateTime.strptime(attrs.delete('mortality_date'), '%m/%d/%y')
+    attrs['spawning_date'] = DateTime.strptime(attrs.delete('spawning_date'), '%m/%d/%y') rescue nil
+    attrs['mortality_date'] = DateTime.strptime(attrs.delete('mortality_date'), '%m/%d/%y') rescue nil
 
     new(attrs)
   end
 
-  validates_inclusion_of :raw, in: [true, false]
```
mdworken: @corneliusellen was the removing of these validations intentional? Looks like they're causing failures in specs (which is good!), so I'd just need to know whether we change the code or change the specs :)

app/controllers/reports_controller.rb
```diff
@@ -5,8 +5,8 @@ def index
 
   def lengths_for_measurement
     @processed_file_id = params[:processed_file_id]
-    @total_animal_lengths = TaggedAnimalAssessment.lengths_for_measurement(@processed_file_id) +
-        UntaggedAnimalAssessment.lengths_for_measurement(@processed_file_id)
+    @total_animal_lengths = TaggedAnimalAssessment.lengths_for_measurement(params[:shl_case_number], params[:date])
+    # + UntaggedAnimalAssessment.lengths_for_measurement(@processed_file_id)
```
mdworken: If we're not using this code anymore I say we just delete it

app/assets/javascripts/charts/growth.js
```diff
@@ -4,36 +4,30 @@ var GrowthChart = function (element, data) {
     title: {
       text: 'Growth Histogram For Measurement ' + data.processed_file_id
     },
-    xAxis: [{
```
mdworken: I did NOT review this file as it's isolated and as the domain expert I trust you know what you want here :)

app/models/tagged_animal_assessment.rb
```diff
@@ -69,8 +69,31 @@ def self.create_from_csv_data(attrs)
     end
   end
 
-  def self.lengths_for_measurement(processed_file_id)
-    select(:length).where(processed_file_id: processed_file_id).map { |record| record.length.to_f }
+  def self.lengths_for_measurement(shl_case_number, measurement_date)
+      measurements = select(:length)
+                     .where(shl_case_number: shl_case_number)
```
mdworken: Super nit picky, but you can actually combine these two `where` clauses into one, not sure which way it reads better though but thought I should just mention :)

PR #122, opened 2020-02-17
Adds DateParser class to prevent silent date parsing errors

### LINKED ISSUE #121: Model Date Parsing Coerces `20**` Years Into `2020`
### Description

The `measurement_date` and `spawning_date` fields of the UntaggedAnimalAssessment and TaggedAnimalAssessment models have unintuitive behavior that is happening silently when dates are not in the exact mm/dd/yy format.

It looks like the reason for these bugs is caused by the `errors.add` code. Because `errors.add` is not being called through a Rails validator, and is instead being called in a custom setter method,  the custom error is not actually being raised. Also, there is no write_attribute in the rescue statement, so if there is an ArgumentError, the setter method will skip trying to set the attribute and will fail silently.

### Examples of Unintuitive Behavior

1. The first example shows how the date fields are accepting incorrectly formatted dates and changing their values in an unexpected way.

#### Steps to Reproduce
```
assessment = FactoryBot.build(:tagged_animal_assessment, measurement_date: "09/08/2019")
```
#### Expected Result
```
assessment.valid? 
=> false
assessment.errors
=> ["Invalid date format: 09/08/2019"]
assessment.measurement_date
=> nil
```
#### Actual Result
```
assessment.valid? 
=> true
assessment.errors
=> []
assessment.measurement_date
=> "2020-09-08"
```

2. The second example shows how the date fields on previously saved rows are not throwing errors when they receive incorrect data.

#### Steps to Reproduce
```
assessment = FactoryBot.create(:tagged_animal_assessment, measurement_date: "09/08/19")
assessment.measurement_date = "invalid"
```
#### Expected Result
```
assessment.save
=> false
assessment.measurement_date
=> nil
```
#### Actual Result
```
assessment.save
=> true
assessment.measurement_date
=> "2019-09-08"
```

colinsoleim:
Resolves #121

### Description

This PR adds a new DateParser class which handles the date parsing for the custom setter methods in TaggedAnimalAssessment and UntaggedAnimalAssessment models. The DateParser class uses a regular expression to ensure the date argument is in the exact correct format to be read by `strptime`. It also gives the user an error message that includes the date format they should be using for these fields.

### Type of change

### How Has This Been Tested?

Steps to reproduce the behavior I reference is included in Issue #121. Several new Rspec tests have been added to test the date validations.

### DIFF
--- a/app/lib/date_parser.rb
+++ b/app/lib/date_parser.rb
@@ -1,12 +1,8 @@
 class DateParser
+  # Returns the parsed date if it's in the mm/dd/yy format, and nil otherwise 
   def self.parse(date)
-    if date_regex.match? date
-      DateTime.strptime(date, "%m/%d/%y")
-    else
-      nil
-    end
-  rescue ArgumentError
-    nil
+    DateTime.strptime(date, "%m/%d/%y") if date_regex.match? date
+    rescue ArgumentError
   end
 
   private
--- a/app/lib/date_parser.rb
+++ b/app/lib/date_parser.rb
@@ -1,0 +1,18 @@
+class DateParser
+  def self.parse(date)
+    if date_regex.match? date
+      DateTime.strptime(date, "%m/%d/%y")
+    else
+      nil
+    end
+  rescue ArgumentError
+    nil
+  end
+
+  private
+
+  # regex to checking mm/dd/yy format for strptime above
+  def self.date_regex
+    Regexp.new('\d{1,2}\/\d{1,2}\/\d{2}\z').freeze
+  end
+end
--- a/app/models/tagged_animal_assessment.rb
+++ b/app/models/tagged_animal_assessment.rb
@@ -51,7 +51,8 @@ class TaggedAnimalAssessment < ApplicationRecord
     spawning_date
   ].freeze
 
-  validates :measurement_date, :shl_case_number, :spawning_date, :tag, :length, presence: true
+  validates :measurement_date, :spawning_date, presence: { message: "must be in the mm/dd/yy format" }
+  validates :shl_case_number, :tag, :length, presence: true
   validates :length, numericality: true
 
   def self.create_from_csv_data(attrs)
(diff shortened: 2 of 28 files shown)


app/lib/date_parser.rb
```diff
@@ -0,0 +1,18 @@
+class DateParser
+  def self.parse(date)
+    if date_regex.match? date
+      DateTime.strptime(date, "%m/%d/%y")
+    else
+      nil
+    end
+  rescue ArgumentError
+    nil
```
wesleyeewong: Do we need these `nil`s? 
 Would this suffice:
```
def self.parse(date)
  DateTime.strptime(date, "%m/%d/%y") if date_regex.match? date
  rescue ArgumentError
end
```

app/models/tagged_animal_assessment.rb
```diff
@@ -51,7 +51,8 @@ class TaggedAnimalAssessment < ApplicationRecord
     spawning_date
   ].freeze
 
-  validates :measurement_date, :shl_case_number, :spawning_date, :tag, :length, presence: true
+  validates :measurement_date, :spawning_date, presence: { message: "must be in the mm/dd/yy format" }
```
wesleyeewong: would it be an issue if we can't determine b/w format errors vs real presence errors?

wesleyeewong:
Cool, didn't even think about doing it that way. 👍

app/lib/date_parser.rb
colinsoleim: Yeah I like that a lot more, a lot less verbose since we're just returning nil in all other cases anyways.

app/models/tagged_animal_assessment.rb
colinsoleim: I was trying to think about this and couldn't come up with any situations where it would matter. I initially thought we could do the presence check and a `validates :format` check, but since we're not saving the value in the `mm/dd/yy` format, it seemed better to me to just have the one check like above.

PR #134, opened 2020-04-18
Add organizations

RachelWyatt:
### Description

This PR adds Organizations to the data structure. The motivation for this is extending the project to the Pinto Abalone organization (and hopefully others in the future!). 

An Organization ```has_many``` Facilities and Users. Facilities and Users, in turn, each ```belong_to``` an Organization. 

This allows for separation of data -- Pinto should not be able to see Facilities or Reports for White Abalone and vice versa.

### Type of change

### How Has This Been Tested?

* Test suite passes with updated factories


app/controllers/facilities_controller.rb
```diff
@@ -69,6 +69,6 @@ def set_facility
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def facility_params
-      params.require(:facility).permit(:name, :code)
+      params.require(:facility).permit(:name, :code, :organization_id)
```
mdworken, line 72: I don't think we want to add this to the params - can't we just calculate it from current_user?

app/controllers/facilities_controller.rb
```diff
@@ -4,7 +4,7 @@ class FacilitiesController < ApplicationController
   # GET /facilities
   # GET /facilities.json
   def index
-    @facilities = Facility.all
+    @facilities = Facility.where(organization_id: current_user.organization_id)
```
jtu0, line 7: How do we feel about always scoping facility to the current user's organization?

I know some people haaaaaaaaaaate default scopes, but I wonder if it wouldn't be a good idea here.

@bhaibel what do you think?

app/models/organization.rb
```diff
@@ -0,0 +1,5 @@
+class Organization < ApplicationRecord
+    has_many :users
+    has_many :facilities
+    has_many :tanks, through: :facilities
```
jtu0, line 4: I _think_ we don't need the `has_many :tanks`, but feel like my understanding of the stakeholder needs is a little too weak to say for sure.

jtu0:
@RachelWyatt want any help getting this PR merged? happy to pair if you need it, or jump in and help on anything you want.

RachelWyatt:
@jtu0 I'll pull down some time this weekend, fix the conflicts and get it merged in! I may slack you if I run into any issues

Issue #139, opened 2020-05-06
Expand "operation" model to allow stateful actions on tanks/families


mdworken:
In order to record the movement of families to/from different tanks, we need to expand/modify our "Operation" model. Let's add attributes to it (or an inheriting sub-class) so that we can record the following actions:

1) Removing a family from a tank.
2) Placing a family in a tank.

**Acceptance criterion:**
- [x] Can remove a family from a tank (by creating a new operation)
- [x] Can add a family to a tank (by creating a new operation)
- [x] Unit + integration tests
- [ ] A deliberate approach to how we'd handle potential "Order of Operations" problems

**Extra credit**

- [ ] Similar, but for the "count" associated with the family/tank. (Will be scoped out more in another ticket later otherwise)

**Out of Scope**

1. Request/System specs
2. Using the CSV uploader

jtu0:
@mdworken Can a family ever be split across multiple tanks?
Or will a family always be in the same tank together?

Similarly, can a tank have multiple families?

jtu0:
- [ ] `Operation` doesn't automatically `.perform` - need to make that happen before closing this issue

mdworken:
Closed via #148

PR #148, opened 2020-05-06
Expand operations to support moving families in and out of tanks

zspencer:
# Checklist:

- [x]  I have performed a self-review of my own code,

See: #139 

### Type of change

### How Has This Been Tested?

Tested by running the rspec tests, there does not appear to be a UI or other interface for performing Operations.


zspencer:
@mdworken - Hey Michael!

If you happen to merge #147 before this, I will rebase it out of this branch; or we can merge this branch as is and close #147; since it's an incidental refactor.

mdworken:
@zspencer I'm so sorry, saw this one second - could you rebase if you don't mind?

zspencer:
@mdworken - Rebase complete!

app/models/operation.rb
```diff
@@ -1,3 +1,25 @@
 class Operation < ApplicationRecord
   belongs_to :tank
+  belongs_to :family, required: false
+
+  validates :family, presence: true, if: :add_family?
```
zspencer, line 5: @jtu0 / @mdworken - Thoughts on this being `if: :needs_family?` to more clearly express intent?

app/models/operation.rb
```diff
@@ -1,3 +1,25 @@
...
+
+  def perform
+    case action.to_sym
```
zspencer, line 8: We call `action&.to_sym` a few different places, we may want to extract the following method:

```
def action
  action&.to_sym
end
```

db/seeds.rb
```diff
@@ -64,7 +64,7 @@
 female = Animal.create!(sex: 'female')
 family = Family.create!(male: male, female: female)
 tank = Tank.create!(facility: Facility.find_by(code: 'PSRF'), name: 'AB-17')
-Operation.create!(tank: tank, animals_added: 800, animals_added_family_id: family.id, operation_date: 7.days.ago, action: 'add')
+Operation.create!(tank: tank, animals_added: 800, family: family, animals_added_family_id: family.id, operation_date: 7.days.ago, action: 'add')
```
zspencer, line 67: @mdworken  - Could you clarify what the purpose of the `animals_added_family_id` is supposed to signify? Is this leftover that can be removed?

spec/models/operation_spec.rb
```diff
@@ -1,7 +1,67 @@
...
+      # syntax does not always play nicely.
+      #
+      # We wrap operation.perform call in a proc so that the lambda expression
+      # can be evaluated by the raise_error block.
+      #
+      # This is part of why some codebases prefer to avoid the "it { ... }" syntax
+      # in favor of "it 'words' do ... end" syntax.
+      subject(:proc) { -> { operation.perform } }
+      let(:action) { :eat_family }
+      it { is_expected.to raise_error(Operation::InvalidActionError) }
```
zspencer, line 64: Another way we could do this without the gross `subject(:proc) { -> { operation.perform } }` would be:

```
it "raises an error" do
  expect { operation.perform }.to raise_error(Operation::InvalidActionError)
end
```

Which is where I _think_ @jtu0 was trying to get us to during the pairing session but I couldn't see the shape of it until now.

db/seeds.rb
mdworken: I think it probably is, yes!

spec/models/operation_spec.rb
mdworken: Eh, I'm fine either way on this honestly.

Issue #140, opened 2020-05-06
Add "OperationEvent" so we can group associated Operations together


mdworken:
Certain Operations might be "related" to each other in terms of business logic. In order to make it easier to expose potential insights we should add a model to track the collection.

**Example use case**: 
GIVEN that I'm a supervisor at a research lab and have multiple researchers
AND different researchers perform functions at different times/on different days
WHEN I view aggregated past data
THEN I'd like to be able to see which actions were performed "together" to investigate if certain researchers/times of day/days of the week are associated with adverse mortality results

**Acceptance Criterion**
- [x]  Add a table OperationEvent that `has_many` Operations
- [x] Give Operations an optional `belongs_to` relationship to OperationEvents
- [x] Specs demonstrating the above

**Out of Scope**
- [ ] Anything actually touching the User models (those will be attached to Operations themselves later)

**Other Notes**
MeasurementEvents are currently done as a joinmodel between Tanks and Measurements - this is NOT a pattern we're going to use going forward.

mdworken:
Addressed in 162. Thanks @zspencer !

PR #162, opened 2020-05-13
✨ Related Operations may be grouped into OperationBatches

zspencer:
See: https://www.github.com/rubyforgood/abalone/issues/140

# Checklist:

- [X] I have performed a self-review of my own code,
- [X] I have commented my code, particularly in hard-to-understand areas,
- [X] I have added tests that prove my fix is effective or that my feature works,
- [X] New and existing unit tests pass locally with my changes ("bundle exec rake"),
- [X] Title include "WIP" if work is in progress.

See #140

### Description
Staff want to be able to group operation(s) in order to understand them at a different level of granularity. (Kind of like how we group related commits into a PR)

### Type of change

### How Has This Been Tested?

Purely through RSpec, as there is no UI for viewing or managing Operations at this time.

### DIFF
--- a/app/models/tank.rb
+++ b/app/models/tank.rb
@@ -5,4 +5,8 @@ class Tank < ApplicationRecord
   has_many :measurements, through: :measurement_events
 
   has_one :family, required: false
+
+  def empty?
+    family.blank?
+  end
 end
--- a/app/models/operation.rb
+++ b/app/models/operation.rb
@@ -1,5 +1,7 @@
 class Operation < ApplicationRecord
   belongs_to :tank
+
+  belongs_to :operation_batch, required: false
   belongs_to :family, required: false
 
   validates :family, presence: true, if: :add_family?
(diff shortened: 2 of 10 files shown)


app/models/tank.rb
```diff
@@ -5,4 +5,8 @@ class Tank < ApplicationRecord
   has_many :measurements, through: :measurement_events
 
   has_one :family, required: false
+
+  def empty?
```
mdworken, line 9: I'm worried about this because `empty?` is a ruby/rails word in of itself. Maybe a synonym would be better?

zspencer: This is an excellent point, Jennifer and I's expectation is that "empty?" is a word we expect Ruby/Rails folks to reach for when attempting to decide if the Tank is empty. _However_ I could see that being an issue if people are using `empty?` as an alternative for `blank?` or `!present?` 

I see `empty?` as useful for collections or things that are not nil, but may not have any contents; while `blank?` is useful for 'this is nil _or_ empty'

```
[].nil? # => false
[].blank? # => true
[].empty? # => true
"".blank?  # => true
"".empty? #=>true
```

LMK what you want us to do, we can take care of it next week! :)

PR #163, opened 2020-05-13
💄Make a gross-sketch version of the operations view ui

zspencer:
We don't currently have a way for people to see the operations that were
performed. This gives us a gross-sketch to start from so they're not
particularly emotionally attached to the layout.

AKA: "Here's a hairy arm"

#  Checklist
- [X] I have performed a self-review of my own code,
- [X] I have commented my code, particularly in hard-to-understand areas,
- [X] I have made corresponding changes to the documentation,
- [X] I have added tests that prove my fix is effective or that my feature works,
- [X] New and existing unit tests pass locally with my changes ("bundle exec rake"),
- [X] Title include "WIP" if work is in progress.

See: https://www.github.com/rubyforgood/abalone/issues/139


zspencer:
@mdworken  / @jtu0  - I was doing my end-of-month 'open pr review' and realized this was failing in CI. I've fixed the lint errors; and am curious if there's more thing you'd like me to do with this or if it's safe to merge?

app/controllers/operations_controller.rb
```diff
@@ -0,0 +1,5 @@
+class OperationsController < ApplicationController
+  def index
+    @operations = Operation.all
```
mdworken, line 3: Let's consider pre-loading here - see the comment below.

app/views/operations/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <th>Family</th>
+  <th>Action</th>
+  <th>CSV Data Uploaded At</th>
+  </tr>
+</thead>
+<tbody>
+<%- @operations.each do |operation|%>
+  <tr>
+    <td><%= operation.tank.name %> </td>
+    <td><%= operation.family&.name %> </td>
```
mdworken, line 14: Would these cause N+1 queries?

zspencer: Yep! Good catch!

app/controllers/operations_controller.rb
zspencer: Done! I think you need both the `includes` and `joins` for this, but I've been known to be wrong. Once or twice. Per hour. Ok fine. Per half-hour.

PR #229, opened 2020-08-09
add usage instructions to the Docker section of the README

robbkidd:
### Description

Some guidance on how to use docker-compose and the make targets.

### Type of change

* Documentation update (_development_)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,33 @@ rake jobs:clear
 ```
 
 ## Docker
-We are currently experimenting with Docker for development. While we would love for more people to try it out be forewarned - Docker functionality may not be maintained moving forward.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,7 +55,33 @@ rake jobs:clear
 ```
 
 ## Docker
-We are currently experimenting with Docker for development. While we would love for more people to try it out be forewarned - Docker functionality may not be maintained moving forward.
+
+We are currently experimenting with Docker for development. While we would love for more people to try it out be forewarned - Docker functionality may not be maintained moving forward. You will need Docker and docker-compose.
```
mdworken, line 59: We might wanna clarify that docker-compose is included in docker desktop.

PR #270, opened 2020-09-05
Fix issue#142

### LINKED ISSUE #142: Associate "Measurement" model to multiple other models
We'd like to measure information about multiple entities in the lab at once. To this end, we'd like to associate the following models to `Measurement`s: `Tank, Animal, Family`. In order to do this we will need to:

- [ ] Sever `has_many through...` association from Tanks -> MeasurementEvent -> Measurement
- [ ] Add the new associations

**Example Use Case**

**Acceptance Criterion**
- [ ] Tank, Animal, Family all `have_many` measurements.
- [ ] Measurements belong_to Tank, Animal, Family
- [ ] The belong_to is optional:true
- [ ] Add appropriate specs
- [ ] Green build (can "Xit" uploader specs if necessary - but explicitly mention it in your PR if you do)

Note: The csv modifications are out of scope and will be addressed in #237

Miguel-Enrique13:
CHANGED models Tank, Family, Animal to reference Measurement.
CHANGED model Measurement to reference Tank, Family, Animal.
CREATED migrations.
CHANGED Tank, Family, Measurement, Animal specs for added references.

Co-authored-by: David Carmo <Dmcarmo@gmail.com>
Co-authored-by: Miguel Guardia <guardia@ualberta.ca>

Resolves #142 

### Type of change

work as expected)
* Documentation update

### DIFF
--- a/app/models/measurement.rb
+++ b/app/models/measurement.rb
@@ -1,6 +1,9 @@
 class Measurement < ApplicationRecord
   belongs_to :measurement_event
   belongs_to :processed_file, optional: true
+  belongs_to :animal, optional: true
+  belongs_to :tank, optional: true
+  belongs_to :family, optional: true
 
   HEADERS = {
     MEASUREMENT_EVENT: "measurement_event",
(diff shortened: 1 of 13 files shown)


app/models/measurement.rb
```diff
@@ -10,20 +13,20 @@ class Measurement < ApplicationRecord
     VALUE: "value"
   }.freeze
 
-  delegate :tank, to: :measurement_event
+  # delegate :tank, to: :measurement_event
```
mdworken, line 16: Let's delete this in the future - I don't think it'll ever come back.

mdworken (approved):
Excellent work, thanks so much for the contribution!

PR #272, opened 2020-09-06
Allow additional models to be associated with measurement through csv upload

### LINKED ISSUE #237: Modify measurement csv uploads to reflect changes from issue #142
We'd like to measure information about multiple entities in the lab at once. To this end, we're going to associate additional models in ticket #142. This issue here is to expand the existing CSV Import to allow the measurement on those additional models to be uploaded.

**Example Use Case**

**Acceptance Criterion**
- [x] Modify the CSV uploader so that we keep current CSV specs green (contact @mdworken for an overview of the architecture there)
- [x] Add appropriate specs
- [x] Any `xit` specs introduced in #142 are now green.

mdworken:
Fixes #237


mdworken:
Closing in deference to #283  (same changes, just squashed and with proper co-authors set)

PR #275, opened 2020-09-06
Install webpacker and Stimulus

### LINKED ISSUE #273: Javascript Framework decisions: Discuss adopting StimulusJS?

benreyn:
Resolves #273


jim (approved):
This looks good to me! Nothing stands out as strange. The real test is adding the first new Stimulus controller, which verifies that all the pieces are working together.

We completely removed Sprockets from Circulate, but I don't think that is necessary to start using webpacker. When we did, though, there were some Rails-specific packages that we had to be added to `package.json` for UJS/Railsy functionality. That might be smoother now (it's been a little while), but it's something to look out for if you make that transition.

benreyn:
Looks like there also a little bit of work to do to get it working on CI. Im also familiar with the "removing sprockets entirely" path. I think we'll head that way eventually, but not quite yet.

benreyn:
To test all this, fire up ye ole' app and check the browser console for `Hello World from Webpacker`.

@robbkidd, tagged your for review specifically for Docker stuff. I am by now means an expert, or even a beginner at Docker stuff, but I tested the updated config locally and it seems to work.

CraigJZ (approved):
I am far from a docker expert, but it looks good to me and is working in both a docker and non-docker setup. I did have to bring in the changes from #280 and #282 locally.

benreyn:
> I did have to bring in the changes from #280 and #282 locally.

Yeah, I did rebase this branch of the main branch. It should work by default. Wondering why you had to manually pull in those changes?

PR #283, opened 2020-09-09
Adds the ability to associate additional models to each Measurement recorded

### LINKED ISSUE #237: Modify measurement csv uploads to reflect changes from issue #142

mdworken:
Fixes #237 

Co-Authored-By: Launa Greer <22844975+LaunaG@users.noreply.github.com>
Co-Authored-By: David <dmbertoglio@gmail.com>
Co-Authored-By: mdworken <31595784+mdworken@users.noreply.github.com>


app/models/family.rb
```diff
@@ -5,7 +5,7 @@ class Family < ApplicationRecord
 
   belongs_to :tank, required: false
 
-  def name
-    "Male: #{male.id} / Female: #{female.id}"
-  end
+  # def name
+  # "Male: #{male.id} / Female: #{female.id}"
+  # end
```
benreyn, line 10: Can we just delete this method? It will be stored in the git history if we ever need it again.

benreyn (approved):

Issue #336, opened 2020-09-19
Scope Tank to an Organization


CraigJZ:
As we plan the multi-tenancy of the application, a Tank should belong to an Organization. 

Acceptance Criteria:

- [x] Add a belongs_to relationship on Tank
- [x] Add a has_many relationship on Organization
- [x] Add an organization_id column to Tank with an index
- [x] Update existing factory for Tank with an Organization
- [x] Add specs to Organization and to Tank for the added relationship

PR #347, opened 2020-09-21
[317] Show Tank Record

### LINKED ISSUE #317: As a user, I can view a Tank record
Acceptance Criteria

- [ ] A new route is added that maps to Tanks#show
- [ ] A new view is added to this route that shows all of the data for a tank
- [ ] A feature test is added that seeds one tank record in the database, loads this page and asserts that content is on the page for that tank

There is some prior art for how we want show pages in our application to look. See Facilities#new and Animals#new. If this task is picked up before #313, Please add to the scope of that issue, or create a new issue for.

- [ ] A link to this page is added to the `Tanks#index` page. 

If this task is picked up after #315, Please create a new ticket to correct the behavior of that form. It should redirect to this page after a new record is created.

smkopp92:
Resolves #317 

### Description
This PR creates a view for the tank show page

Include anything else we should know about. -->
From Power Abalone Create team

### Type of change

### How Has This Been Tested?

After updating and re-running the seed file, you can visit a tanks show page through the URL (i.e. http://localhost:3000/tanks/4). A feature test has also been created for this page.

### Screenshots
Screen Shot 2020-09-21 at 1 20 30 PM

### DIFF
--- a/app/controllers/tanks_controller.rb
+++ b/app/controllers/tanks_controller.rb
@@ -1,0 +1,11 @@
+class TanksController < ApplicationController
+  before_action :set_tank, only: [:show]
+
+  def show; end
+
+  private
+
+  def set_tank
+    @tank = Tank.find(params[:id])
+  end
+end
--- a/spec/features/tanks/show_spec.rb
+++ b/spec/features/tanks/show_spec.rb
@@ -1,0 +1,18 @@
+require 'rails_helper'
+
+describe "When I visit the tank Show page" do
+  let(:user) { create(:user) }
+
+  before do
+    sign_in user
+  end
+
+  it "Then I see information of a specific tank" do
+    tank = FactoryBot.create(:tank)
+
+    visit tank_path(tank)
+
+    expect(page).to have_content(tank.name)
+    expect(page).to have_content(tank.facility.name)
+  end
+end
--- a/app/views/tanks/show.html.erb
+++ b/app/views/tanks/show.html.erb
@@ -1,0 +1,20 @@
+<section class="section facilities__show">
+  <div class="container facilities__show__container">
+    <p id="notice"><%= notice %></p>
+    <div class="facilities__show__container__header">
+      <h1 class="title">Tank</h1>
+    </div>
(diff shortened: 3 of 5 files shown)


smkopp92:
@CraigJZ @benreyn For this tank show page PR, we know that there are other relationships to models that we have not included yet, such as post-settlement-inventories and measurements. Should we focus on adding these relationships to this story/issue, or is there a separate issue where these relationships will be tackled?

spec/features/tanks/show_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    sign_in user
+  end
+
+  it "Then I see information of a specific tank" do
+    tank = FactoryBot.create(:tank)
+
+    visit tank_path(tank)
+
+    expect(page).to have_content(tank.name)
+    expect(page).to have_content(tank.facility.name)
```
benreyn, line 16: Thoughts on adding a delegation here to help (somewhat) mitigate [LOD](https://en.wikipedia.org/wiki/Law_of_Demeter) violations here and in the future?

app/controllers/tanks_controller.rb
```diff
@@ -0,0 +1,11 @@
+class TanksController < ApplicationController
+  before_action :set_tank, only: [:show]
+
+  def show; end
+
+  private
+
+  def set_tank
+    @tank = Tank.find(params[:id])
```
benreyn, line 9: After #336 is implemented we will want to follow up and make sure that we scope any queries here to the `current_user`'s organization, but fine to leave as is for now I think

benreyn (approved):
Just a few notes here, but nothing that should be considered a blocker for merging. 🥇 

Thanks for tackling this so quickly! 🎉

spec/features/tanks/show_spec.rb
smkopp92: I like that idea, something like a `facility_name`?

benreyn: Yup, I was thinking something along the lines of...

```ruby
delegate :name, to: :facility, prefix: true
```

on the `Tank`

Issue #320, opened 2020-09-19
As a user, I can delete a Family record


benreyn:
Acceptance Criteria

- [ ] A route is added that maps to `Families#destroy`
- [ ] A button/link is added to the `Families#index` page created in #312
- [ ] This action should ask for confirmation before this destructive action is taken
- [ ] A feature test is added that seeds a family record in the database, loads the `Families#index` page and clicks to delete the record, then asserts the record is no longer on the page

Issue #318, opened 2020-09-19
As a user, I can update a Family record


benreyn:
Acceptance Criteria

- [ ] A new route is added that maps to Families#edit
- [ ] A new view is added that renders a form, pre-populated with the data for the existing Family record, for the user to edit. Idealy this could re-use the markup for the form created in #314 
- [ ] This form can be submitted to update the family record in the database.
- [ ] A feature test is added that seeds the database with a family record, loads this page, makes an edit to that record, and submits the form

Issue #316, opened 2020-09-19
As a user, I can view a Family record


benreyn:
Acceptance Criteria

- [ ] A new route is added that maps to Family#show
- [ ] A new view is added to this route that shows all of the data for a family
- [ ] A feature test is added that seeds one family record in the database, loads this page and asserts that content is on the page for that family

There is some prior art for how we want show pages in our application to look. See Facilities#new and Animals#new. If this task is picked up before #312, Please add to the scope of that issue, or create a new issue for.

- [ ] A link to this page is added to the `Families#index` page. 

If this task is picked up after #314, Please create a new ticket to correct the behavior of that form. It should redirect to this page after a new record is created.

PR #348, opened 2020-09-21
Add CRUD pages for family resource

### LINKED ISSUE #312: As a user, I can view a list of all families in my organization
Acceptance Criteria

- [ ] A new route is added that maps to `Families#index`
- [ ] A new view is added that lists all `Family` records. This list should be scoped to the organization that the `current_user` `belongs_to`
- [ ] A link to this page is added to the navigation bar
- [ ] A basic "smoke test" is added that seeds that database with a few family records (probably 2), loads the page and asserts that these records exist on the page.

There is some prior art for how we want index pages in our application to look. See `Facilities#index` and `Animals#index`. Some of the links that will exist on this page, (new, show, edit, delete). Its fine to leave this blanks and fill those links in later as a separate task, or as a part of the task to add those actions.

todtb:
#

Resolves #312, #314, #316, #318, #320    

### Description
Adds CRUD pages for `Family` resource. 

### Type of change

### How Has This Been Tested?

A few feature specs included. New resource routes for `Family` can be navigated to in order to verify.

### DIFF
--- a/app/views/families/index.html.erb
+++ b/app/views/families/index.html.erb
@@ -1,0 +1,39 @@
+<section class="section facilities__index">
+  <div class="container facilities__index__container">
+    <p id="notice"><%= notice %></p>
+
+    <div class="facilities__index__container__header">
+      <h1 class="title">Families</h1>
+      <%= link_to new_family_path, class: "button-primary" do %>
+        New Family <%= image_tag  "icons/plus.png", size: 17 %>
+      <% end %>
+    </div>
+
+    <table class="table  table-hover table-responsive-sm">
+      <thead>
+      <tr class="bg-dark-light">
+        <th>Name</th>
+        <th>Female</th>
+        <th>Male</th>
+        <th>Tank</th>
+        <th colspan="3"></th>
+      </tr>
+      </thead>
+
+      <tbody>
+      <% @families.each do |family| %>
+        <tr>
+          <td><%= family.name %></td>
+          <td><%= link_to family.female_pii_tag, animal_path(family.female) %></td>
+          <td><%= link_to family.male_pii_tag, animal_path(family.male) %></td>
+          <td><%= family.tank&.name %></td>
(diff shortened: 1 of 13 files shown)


app/views/families/index.html.erb
```diff
@@ -0,0 +1,39 @@
+<section class="section facilities__index">
+  <div class="container facilities__index__container">
+    <p id="notice"><%= notice %></p>
```
benreyn, line 3: I noted this in #349 but we should make a follow up task to cleanup our usage of `notice` and `altert` and convert to `flashes` as a hash. (no action needed here)

app/views/families/index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <th colspan="3"></th>
+      </tr>
+      </thead>
+
+      <tbody>
+      <% @families.each do |family| %>
+        <tr>
+          <td><%= family.name %></td>
+          <td><%= link_to family.female.pii_tag, animal_path(family.female) %></td>
+          <td><%= link_to family.male.pii_tag, animal_path(family.male) %></td>
```
benreyn, line 28: Thoughts on adding delagation to `Family` that will give us `male_pii_tag` and `female_pii_tag` methods on family?

benreyn (approved):
Looks good! Way to close 5 issues in one PR @todtb !

app/views/families/index.html.erb
todtb: Sounds good @benreyn. This is going to be the primary identifier for end user use, right? What would the `tag_id` be?

todtb:
Hey @benreyn, changes made, just pending one last co-author confirmation of user/email which I'll amend before we continue.

app/views/families/index.html.erb
benreyn: Im working on teasing that out from the data we have from our stakeholders, and digging through legacy code from when I joined the project. Ill follow up on this ASAP.

PR #349, opened 2020-09-21
[313] Added an index view for tanks

### LINKED ISSUE #313: As a user, I can view a list of all tanks in my organization
Acceptance Criteria

- [ ] A new route is added that maps to `Tanks#index`
- [ ] A new view is added that lists all `Tank` records. This list should be scoped to the organization that the `current_user` `belongs_to`
- [ ] A link to this page is added to the navigation bar
- [ ] A basic "smoke test" is added that seeds that database with a few tank records (probably 2), loads the page and asserts that these records exist on the page.

There is some prior art for how we want index pages in our application to look. See `Facilities#index` and `Animals#index`. Some of the links that will exist on this page, (new, show, edit, delete). Its fine to leave this blanks and fill those links in later as a separate task, or as a part of the task to add those actions.

michellemhey:
# Checklist:

Resolves #313

### Description
This PR builds an index page for tanks.

### Type of change
New feature

### How Has This Been Tested?
Wrote a feature spec and tested that the page works locally.

### Screenshots
Screen Shot 2020-09-21 at 2 46 35 PM

### DIFF
--- a/app/controllers/tanks_controller.rb
+++ b/app/controllers/tanks_controller.rb
@@ -1,6 +1,10 @@
 class TanksController < ApplicationController
   before_action :set_tank, only: [:show]
 
+  def index
+    @tanks = Tank.all
+  end
+
   def show; end
 
   private
--- a/app/models/tank.rb
+++ b/app/models/tank.rb
@@ -8,7 +8,7 @@ class Tank < ApplicationRecord
   has_one :family, required: false
 
   delegate :name, to: :facility, prefix: true
-  
+
   def empty?
     family.blank?
   end
--- a/app/models/tank.rb
+++ b/app/models/tank.rb
@@ -7,6 +7,8 @@ class Tank < ApplicationRecord
 
   has_one :family, required: false
 
+  delegate :name, to: :facility, prefix: true
+  
   def empty?
     family.blank?
   end
--- a/app/views/tanks/index.html.erb
+++ b/app/views/tanks/index.html.erb
@@ -1,0 +1,34 @@
+<section class="section facilities__index">
+  <div class="container facilities__index__container">
+    <p id="notice"><%= notice %></p>
+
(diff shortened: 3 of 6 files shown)


app/controllers/tanks_controller.rb
```diff
@@ -1,6 +1,10 @@
 class TanksController < ApplicationController
   before_action :set_tank, only: [:show]
 
+  def index
+    @tanks = Tank.all
```
benreyn, line 5: Similarly with #347, we will want to circle back to scoping the queries in this controller once #336 is completed, but that doesnt need to be a blocker to getting this work merged. we can make a follow up issue.

app/models/tank.rb
```diff
@@ -7,6 +7,8 @@ class Tank < ApplicationRecord
 
   has_one :family, required: false
 
+  delegate :name, to: :facility, prefix: true
```
benreyn, line 10: I ❤️ this! Thanks for adding it in this PR.

app/views/tanks/index.html.erb
```diff
@@ -0,0 +1,34 @@
+<section class="section facilities__index">
+  <div class="container facilities__index__container">
+    <p id="notice"><%= notice %></p>
```
benreyn, line 3: Im just now "noticing" 😏 That we render `notice` both here and in the application layout [here](https://github.com/rubyforgood/abalone/blob/master/app/views/layouts/application.html.erb#L102-L107). We should make a follow up issue to unify these so that its only need in one place, and probably use `flashes` instead of `notice` and `alert`. 

Not an action item here, but wanted to "note" it so it didnt get forgotten.

benreyn (approved):
A few comments about follow up tasks we will need to make, but nothing here is a blocker to merging this work. Thanks @michellemhey !

Issue #334, opened 2020-09-19
Scope Family to an Organization


CraigJZ:
As we plan the multi-tenancy of the application, a Family should belong to an Organization.

Acceptance Criteria:

- [x] Add a belongs_to relationship on Family
- [x] Add a has_many relationship on Organization
- [x] Add an organization_id column to Family with an index
- [x] Update existing factory for Family with an organization
- [x] Add specs to Organization and to Family for the added relationship

haydenrou:
In the family CRUD, we have used `Family.all` to show families on the index page - we'll want to make use of a scope here per organisation, instead of showing all when this task is being done.

PR #351, opened 2020-09-21
Add organization column to [Tank, Family, Animal]

### LINKED ISSUE #333: Scope Animal to an Organization
As we plan the multi-tenancy of the application, a Animal should belong to an Organization.

Acceptance Criteria

- [x] Add a belongs_to relationship on Animal
- [x] Add a has_many relationship on Organization
- [x] Add an organization_id column to Animal with an index
- [x] Update existing factory for Animal with an organization
- [x] Add specs to Organization and to Animal for the added relationship
- [x] Scope queries that occur on the `AnimalsController` to the current logged in users organization

haydenrou:
Add organization association to animal
Retofit specs to accommodate organizations for the csv importer.
Update the animals controller to scope using the new
Add a concern for OrganizationScope which can be added to any model with
an organization association

Resolves #333 #334 #336

### Type of change

work as expected)
* Documentation update


haydenrou:
Hi @benreyn we are looking to address #336 #334 #333 in this PR. We have only started #333 here and wouldn't mind some feedback on if you are happy with the approach we are taking before pushing the rest up.

As we are adding foreign keys here, we'd be interested to know if these tables are being used yet, so as whether we need to write a back-fill migration to fill the `organization_id`'s in for each of these models.

benreyn:
> we'd be interested to know if these tables are being used yet, so as whether we need to write a back-fill migration to fill the organization_id's in for each of these models. 

No need for this, as we haven't reached our MVP yet, we are safely able to nuke the database and start fresh. One thing to note though is that the db:seed tasks might need to be updated.

> Additionally, would you be opposed to us adding organization_id to many other models? We were thinking of streamlining this across the application's models more than just the three aforementioned ones.

Curious which models you have in mind. (note that many of the db tables are actually dead code that we need to eventually remove, See [these](https://github.com/rubyforgood/abalone/issues?q=is%3Aissue+is%3Aopen+label%3A%22Technical+Debt%22) issues if your curious about which ones.

The only models that we will need to account for this for in the MVP are `Tank`, `Facility`, `Animal`, `Family`, `Measurement`, `Operation`, `Measurement` and `MeasurementEvent`.

@CraigJZ Did I miss anything there?

@haydenrou, if there are models that we have missed in that list above, then yes, we do want to make sure that they are also organization level scoped.

CraigJZ:
@benreyn No, I think that covers it. Thanks!!

app/controllers/animals_controller.rb
```diff
@@ -2,7 +2,7 @@ class AnimalsController < ApplicationController
   before_action :set_animal, only: [:show, :edit, :update, :destroy]
 
   def index
-    @animals = Animal.all
+    @animals = Animal.for_organization(current_user.organization)
```
haydenrou, line 5: `current_organization`

CraigJZ (commented):
Sorry for the delay in getting to this. I think we're good here but just need to resolve the merge conflict.

Thanks!!!!

todtb (approved):

PR #356, opened 2020-09-22
[338] Feature - Added role column to users table

### LINKED ISSUE #338: Introduce the concept of an admin user at the organization level
An organization has users. An admin user should have all the access of a regular user to an organization plus to able to view, add, update and destroy other users belonging that that organization.

There is the potential for some additional roles at the organizational level in the future but we are only concerned with user and admin right now. A simple, lightweight approach would be best here. The 3rd option (Using Active Record enum) presented [here](https://github.com/heartcombo/devise/wiki/How-To:-Add-an-Admin-Role) should be a good choice.

Acceptance Criteria

- [ ] Introduce the concept of an admin user on the User model following the guidance here: https://github.com/heartcombo/devise/wiki/How-To:-Add-an-Admin-Role
- [ ] Add the enum column on User
- [ ] Add specs for checking the role of a user

michellemhey:
# Checklist:

Resolves #338 

### Description
This PR adds the role column to the users table and sets a default role of 'user' when a user is created.

### Type of change
New feature

### How Has This Been Tested?
Added a model spec and created a new user record in the DB to see if a default role was set.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,5 @@
 class User < ApplicationRecord
-  enum role: [:user, :admin]
+  enum role: %i[user admin]
   after_initialize :set_default_role, if: :new_record?
 
   def set_default_role
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,11 @@
 class User < ApplicationRecord
+  enum role: [:user, :admin]
+  after_initialize :set_default_role, if: :new_record?
+
+  def set_default_role
+    self.role ||= :user
+  end
+
   # Include default devise modules. Others available are:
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -1,4 +1,11 @@
 class User < ApplicationRecord
+  enum role: %i[user admin]
+  after_initialize :set_default_role, if: :new_record?
+
+  def set_default_role
+    self.role ||= :user
```
benreyn, line 6: I prefer encapsulating creation logic in a service object over using ActiveRecord callbacks, as the callback route has bitten me in the past, but its fine to go this route for now as its only one thing we piece of logic we need to add at the time of record creation.

benreyn (approved):

PR #359, opened 2020-09-22
Use alpine and fix yarn install

iblackman:
### Description
Please also include relevant motivation and context.
Guide questions:
  - What motivated this change (if not already described in an issue)?
    We were not able to start up the docker environment using `docker-compose`. It was throwing an error about webpacker. Since the dockerfile was already needing some attention we decided to change it a bit.
  - What alternative solutions did you consider?
    Trying to fix with the old dockerfile didn't look too promising.
  - What are the tradeoffs for your solution?
    It works now.
   
* No changes.

### Type of change

### How Has This Been Tested?

Just follow up the README instruction for the docker environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,16 @@
 FROM ruby:2.6.6-alpine AS builder
 
-LABEL maintainer="jeanine@littleforestconsulting.com"
+ARG RAILS_ROOT=/usr/src/app/
+WORKDIR $RAILS_ROOT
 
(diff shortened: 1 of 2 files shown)


bcdonadio (approved):
LGTM

benreyn (approved):
Thanks for giving our docker setup some love @iblackman ! Docker is really import for the ability for contributors on Windows machines to be productive! Thanks!

PR #370, opened 2020-09-22
Add animal csv uploads

### LINKED ISSUE #305: CSV upload for Animal
As we onboard new organizations to our software, we will need a way for them to import their existing data in large batches. The best way to accomplish this is with a CSV upload. We should provide them with a page that allows them to...

- [ ] Download a CSV file that specifies the format that they should upload in
- [ ] Upload a CSV that matches this specified format
- [ ] Inform them of the status of this upload
  - completed (all records uploaded)
  - failed (some kind of error occured)
  - queue (has not started processing yet)
  - processing

This ticket encompasses only the CSV based upload for Animal records.

We have some existing CSV upload infrastructure, but I wouldn't let that guide you too much on this issue, as that functionality was built for an earlier iteration of this application and may not longer be the best way to handle CSV based data-upload.

todtb:
Adds ActiveStorage and a new FileUpload model to track user,
organization, status, and attachment. Files are uploaded, stored using
ActiveStorage and then processed in a DelayedJob. This job creates an
animal provided it follows the new composite uniqueness constraint on PII
tag and organization.

Co-authored-by: Hayden Rouille <hayden@rouille.net>
Co-authored-by: Vinicius Gama <vinicius.sgama@gmail.com>
Co-authored-by: Ayako Kurushima <ayakokurushima@gmail.com>
Co-authored-by: Sebastian Kopp <smkopp92@gmail.com>

Wanted some feedback before continuing, therefore specs/relevant docs pending.

Resolves #305  

### Type of change

### How Has This Been Tested?

Locally by uploading a few CSVs (some with duplicate or invalid data) and verifying non-duplicate records are created.


app/models/file_upload.rb
```diff
@@ -0,0 +1,7 @@
+class FileUpload < ApplicationRecord
```
todtb, line 1: I can see this replacing `processed_files` and `temporary_files` using ActiveStorage instead of serializing the file contents.

app/jobs/animals_job.rb
```diff
@@ -0,0 +1,40 @@
+class AnimalsJob < ApplicationJob
```
todtb, line 1: Still going to used `delayed_job` to parse this all, output can be stored in new table.

app/models/file_upload.rb
```diff
@@ -0,0 +1,7 @@
...
+  belongs_to :user
```
todtb, line 2: It would be nice to add a quick new feature for a simple page showing a users uploads along with the status (if the job is pending, if there were errors), as the data is all there now.  This would replace the existing Uploads > File List

app/jobs/animals_job.rb
smkopp92, line 1: Do we want to be a little more specific with the name of this job? Maybe something like `ImportAnimalsJob`?

todtb: Good call

spec/features/animals/upload_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
```
smkopp92, line 1: Good work with these specs!

smkopp92 (approved):
LGTM! We should consider now removing the outdated csv upload pattern so that we don't get confused of which route to take for uploads in the future

haydenrou (approved):

PR #371, opened 2020-09-22
[354] Make animals.sex an enum

### LINKED ISSUE #354: Make animals.sex an enum
The `sex` column on the `animals` table is currently stored as a raw string with no validation. To help ensure our data-integrity long term we should migrate this to an `enum` column. My preference would be to use a [native Postgres enum type](https://dev.to/amplifr/postgres-enums-with-rails-4ld0), as we will almost certainly want to query this data in SQL one day, and mapping rails `enum`s to integers in the database leads to more back and forth than I tend to like. Two gems that make this process a little easier are [activerecord-postgres_enum](https://github.com/bibendi/activerecord-postgres_enum) and [activerecord-pg_enum](https://github.com/alassek/activerecord-pg_enum). Feel free to use or not use either of these gems.

Worth noting here is that based on what I've been able to tell from the sample data files we've gotten from our stakeholders, it appears that they are not always able to determine the sex of a given animal (my assumption here is that this is due to maturity?). This can either be represented by a `null` value in the `sex` column, or (preferably IMO) a third `unknown` option for this enum type.

As we dont have any data we need to preserve, we can ignore migrating old data to the new column as a part of this task.

viniciusgama:
# Checklist:

Resolves #354 

### Description
  - _What motivated this change (if not already described in an issue)?_ Normalize the data round the sex of an animal. Currently this is a open text field which can be difficult to query in the future if different users inputs different values.
  - _What alternative solutions did you consider?_ The current solution which can led to future problems.
  - _What are the tradeoffs for your solution?_ Adds an extra dependency

This also make sure that when inputing the data in the form, the user can only select valid options.

**_List any dependencies that are required for this change. (gems, js libraries, etc.)_**

[activerecord-postgres_enum](https://github.com/bibendi/activerecord-postgres_enum)

I have chosen this option as it's the one with the most recent releases which indicates it's being actively being developed

### Type of change

### Screenshots

Screen Shot 2020-09-22 at 7 10 50 PM


app/views/animals/show.html.erb
```diff
@@ -32,7 +32,7 @@
 
       <p>
         <strong>Sex:</strong>
-        <%= @animal.sex %>
+        <%= @animal.sex&.humanize %>
```
haydenrou, line 35: Is there ever a scenario where there is no sex?

haydenrou (approved):

app/views/animals/show.html.erb
viniciusgama: not sure. I've added this here because the field is nullable. Should we restrict this @benreyn @CraigJZ  ?

benreyn: Actually yes! In our meeting with stakeholders yesterday, and in the data files they have sent us, there are many cases where the sex of an animal is unknown or not able to be determined until the animal is more mature. 

My preference would be to make the column non-nullable in the database and add a third "unknown" option to the enum.

viniciusgama: implemented here @haydenrou @benreyn 8631e19ab21ab034bc2c8800347082d18966dac0

benreyn (approved):
Awesome! Thanks @viniciusgama !

PR #372, opened 2020-09-23
Include facilities for tanks to avoid n+1 query

### LINKED ISSUE #365: N+1 query
There’s a N+1 issue with facilities when loading the tanks page. Worth a PR to stamp that out. We should be checking for that on any pages we’re contributing. (wade)

haydenrou:
This PR aims to fix an n+1 query on the tanks page.

# Checklist:

-->

Resolves #365 

### Type of change

work as expected)
* Documentation update

### DIFF
--- a/app/controllers/tanks_controller.rb
+++ b/app/controllers/tanks_controller.rb
@@ -2,7 +2,7 @@ class TanksController < ApplicationController
   before_action :set_tank, only: [:show, :edit, :update, :destroy]
 
   def index
-    @tanks = Tank.for_organization(current_organization)
+    @tanks = Tank.for_organization(current_organization).includes(:facility)
   end
 
   def show; end


app/controllers/tanks_controller.rb
```diff
@@ -2,7 +2,7 @@ class TanksController < ApplicationController
   before_action :set_tank, only: [:show, :edit, :update, :destroy]
 
   def index
-    @tanks = Tank.for_organization(current_organization)
+    @tanks = Tank.for_organization(current_organization).includes(:facility)
```
viniciusgama, line 5: any way we can solve this for every model that includes the `OrganizationScope` and uses the `for_organization`? This way we can avoid spreading `includes` for every usage this method.

benreyn (approved):
Awesome! Thanks for getting this fixed @haydenrou !

app/controllers/tanks_controller.rb
haydenrou: The `includes` in this scenario is for the facility, so I wouldn't be sure that this would be overly relevant for everywhere the organization is being used, however if your concern was an N+1 for the organization, we could add an `includes` to the scope.

As for defaulting the scope, we discussed placing this `for_organization` scope as a default, however it's a bit hacky getting the `current_user` into the model. I'd suggest perhaps a better way would be to use `scopes` from [the pundit gem](https://github.com/varvet/pundit) giving potentially a nicer way to access "default" or authorized access to data. This way you could put `policy_scope(Tank)` into the controller.

Maybe this ^^ is a nicer way than the default scope @wadewinningham @todtb that we were briefly discussing yesterday?

PR #376, opened 2020-09-23
[340] Feature - Returns a list of users

### LINKED ISSUE #340: As an admin, I can see a list of all users for an organization
- [ ] The UsersController index action is updated to return all users for the organization that the current_user belongs_to
- [ ] A new view is added that lists all users for an organization
- [ ] A link to this page is added to the address bar with a condition that it is only visible to admin
- [ ] A basic test is added that seeds that database with a few user records (probably 2), loads the page and asserts that these records exist on the page.

There is some prior art for how we want index pages in our application to look. See Facilities#index and Animals#index.

This task is dependent on #339

michellemhey:
# Checklist:

Resolves #340

### Description
This PR creates the index view for users. Only users with an admin role can view the index page and the nav link.

### Type of change
New feature

### How Has This Been Tested?
There are feature and controller specs. I manually tested it as well


CraigJZ (approved):
Looks great! Thank you!

PR #375, opened 2020-09-23
[328] Remove post settlement inventory model and related code

miniyakkos:
Resolves #328 

### Description
Removes PostSettlementInventory model and related code.

### Type of change
* Refactoring

### How Has This Been Tested?
Ran entire test suite

PR #379, opened 2020-09-23
Rework measurements to have correct associations and values

### LINKED ISSUE #308: Measurement / MeasurementType Interface
An organization needs to define an acceptable list of measurement names and measurement value types for measurements that will be added to the system. This will most likely take the form of an interface exposed at the organization level.

Examples:
```
Measurement Name: Length
Measurement Value Type: Centimeters

Measurement Name: Age
Measurement Value Type: Days

Measurement Name: Weight
Measurement Value Type: Grams

Measurement Name: Quantity
Measurement Value Type: Units
```

The current structure of the Measurement model points to an approach where the acceptable measurement names and measurement value types are configured for the organization and then used to validate measurements as they are accepted into the system.

An alternative approach being considered is to add a MeasurementType model with the name, value type and an organization_id. A Measurement belongs_to a MeasurementType and a MeasurementType has_many measurements. A MeasurementType belongs_to an organization and an Organization has_many measurement types. The acceptable list would be the MeasurementType records for an organization and those would be used for Measurements being accepted into the system.

#### Tasks: 

Note: These are broad and can be broken out into individual issues

- [x] Determine an approach
- [x] Add the MeasurementType model and relationships, if going with this approach
- [x] Update the relevant seed files, factories and specs
- [x] Build out the interface for defining these values at the organization level.

Note: The concept of an "organization level" means that an admin user for an organization would be defining these values. If this issue is addressed before #338, this concept can be ignored with the understanding that we would scope that level of access once #338 has been completed.

haydenrou:
Add a measurement type model for tracking different measurements per
organization for different values (e.g. length, count) and units (e.g. cm,
integer)
Move around associations to have a polymorphic assocations for the test
subject, and measurements, measurement types and measurement events to
belong to each other correctly.
Remove tank_id from measurement event as they could belong to many
different things which you can diverge from the associated measurements.
Add specs for all of the above.
Deprecate specs for CSVs and File Uploads, and assocaited methods on
models, such as #create_from_csv . The reason for this is that they are
going to be entirely reworked, and the current usage of them only
facilities specs, and not provides functionality to the app. I'm
expecting all these skipped tests will be removed further down the line
when the CSV import is reworked.

Co-authored-by: Brian Todt <todtb@users.noreply.github.com>
Co-authored-by: Sebastian Kopp <smkopp92@gmail.com>

# Checklist:

- [ ] I have performed a self-review of my own code,
- [ ] I have commented my code, particularly in hard-to-understand areas,
- [ ] I have added tests that prove my fix is effective or that my feature works,
- [ ] New and existing unit tests pass locally with my changes ("bundle exec rake"),
- [ ] Title include "WIP" if work is in progress.

Resolves #308 

### Type of change

work as expected)
* Documentation update


spec/models/tank_spec.rb
```diff
@@ -4,6 +4,7 @@
   it "Tank has associations" do
     is_expected.to belong_to(:facility).optional
     is_expected.to belong_to(:organization)
+    is_expected.to have_many(:post_settlement_inventories)
```
benreyn, line 7: This can safely be removed. `PostSettlementInventory` is one of the old models that was removed in #375

spec/factories/measurements.rb
```diff
@@ -1,9 +1,9 @@
 FactoryBot.define do
   factory :measurement do
-    name { "MyString" }
-    value_type { "MyString" }
     value { "" }
+    measurement_type
+    association :subject, factory: :tank
```
benreyn, line 5: I think `:animal` will likely end up being the default we want here, and we will likely add traits for the rest of the possibilities in the polymorphic association.

db/seeds.rb
```diff
@@ -77,6 +77,10 @@
...
-Measurement.create!(name: 'count', value_type: 'integer', value: '719', measurement_event: measurement_event, date: 1.days.ago, organization_id: white_abalone.id)
+measurement_event = MeasurementEvent.create!(name: "My first measurement", organization_id: white_abalone.id)
+measurement_type = MeasurementType.create!(name: "length", unit: "cm", organization: white_abalone)
+measurement_type = MeasurementType.create!(name: "count", unit: "number", organization: white_abalone)
+measurement_type = MeasurementType.create!(name: "weight", unit: "g", organization: white_abalone)
+measurement_type = MeasurementType.create!(name: "gonad score", unit: "number", organization: white_abalone)
+Measurement.create!(value: '743', measurement_event: measurement_event, date: 3.days.ago, organization_id: white_abalone.id, subject: male, measurement_type: measurement_type)
+Measurement.create!(value: '719', measurement_event: measurement_event, date: 1.days.ago, organization_id: white_abalone.id, subject: female, measurement_type: measurement_type)
```
benreyn, line 86: I think this is fine, but we will probably need to revisit these seeds at some point in the future to make sure that it sets up the application  in a state that accurately portrays its daily usage.

db/schema.rb
```diff
@@ -106,33 +106,37 @@
...
-    t.index ["family_id"], name: "index_measurements_on_family_id"
+    t.string "subject_type", null: false
+    t.bigint "subject_id", null: false
+    t.bigint "measurement_type_id"
     t.index ["measurement_event_id"], name: "index_measurements_on_measurement_event_id"
+    t.index ["measurement_type_id"], name: "index_measurements_on_measurement_type_id"
     t.index ["organization_id"], name: "index_measurements_on_organization_id"
     t.index ["processed_file_id"], name: "index_measurements_on_processed_file_id"
-    t.index ["tank_id"], name: "index_measurements_on_tank_id"
+    t.index ["subject_type", "subject_id"], name: "index_measurements_on_subject_type_and_subject_id"
```
benreyn, line 139: Good call adding this index 👍

db/schema.rb
```diff
@@ -10,15 +10,15 @@
...
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
   create_enum :animal_sex, [
-      "unknown",
-      "male",
-      "female",
+    "unknown",
+    "male",
+    "female",
```
benreyn, line 21: Weird that this ended up in the diff. Maybe it has to do with the version of Postgres your running locally? Ive seen that be the cause of weird diffs in `schema.rb` in other projects in the past.

app/models/tank.rb
```diff
@@ -3,9 +3,8 @@ class Tank < ApplicationRecord
 
   belongs_to :facility, optional: true
   has_many :operations, dependent: :destroy
-  has_many :measurement_events, dependent: :destroy
-  # has_many :measurements, through: :measurement_events
-  has_many :measurements
+  has_many :post_settlement_inventories
```
benreyn, line 6: Safe to remove.

app/models/measurement_type.rb
```diff
@@ -0,0 +1,7 @@
+class MeasurementType < ApplicationRecord
+  include OrganizationScope
```
benreyn, line 2: I dont know when this got added, but I ❤️ this abstraction.

benreyn (commented):
This is a HUGE step forward for us! Thanks for taking this on @haydenrou, @todtb, and @smkopp92 !
Left a few comments but nothing too huge! This pretty close to being able to merge. 

I hate having this much work in one commit, but I think it will be best to squash and merge this one, if you dont mind. We'll be iterating on things anyway as the CSV stuff needs to get reworked.

db/seeds.rb
haydenrou: I agree. I've felt like we have slightly been writing these tests to pass from a function perspective but not necessarily use case perspective. Shall we add a backlog task to review the seeds file and make some amendments where seen fit? For instance I think both organizations should have measurements, not just white_abalone like this.

CraigJZ (approved):

benreyn (approved):

db/migrate/20200923150158_create_measurement_types.rb
```diff
@@ -0,0 +1,11 @@
+class CreateMeasurementTypes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :measurement_types do |t|
+      t.string :name # count || length
+      t.string :unit # integer || cm
+      t.references :organization
```
NickSchimek, line 6: should there be a unique index here `t.index [:name, :organization]`

todtb: Possibly, I suppose it is possible to have a measurement performed in two units, so I guess some clarification around how it's used. If, say, "length" were always "cm" then yes, I think you could have one uniquely named measurement type per org. If there could be "length" - "cm", "length" - "mm" that would be my concern.

NickSchimek: That makes sense. Thanks for clarifying. Perhaps there should be a unique index for all three columns?

todtb: Yeah, that would almost certainly be safe to assume in my opinion, and the db level constraint is always a nice extra safeguard.

app/models/measurement.rb
```diff
@@ -16,23 +18,17 @@ class Measurement < ApplicationRecord
...
-        organization_id: attrs.fetch(:organization_id)
-      )
-    end
-    family = Family.find_by!(name: attrs.fetch(:family_name)) if attrs[:family_name]
+    tank = Tank.find_or_create_by!(
+      name: attrs.fetch(:tank_name),
+      organization_id: attrs.fetch(:organization_id)
+    )
+
+    measurement_type = MeasurementType.find_or_create_by!(name: "length", unit: "cm", organization_id: attrs.fetch(:organization_id))
```
mdworken, line 31: I'm having a little trouble understanding this - it looks like it hardcodes the incorrect measurement type. Was there supposed to be a follow-on commit?

app/models/measurement.rb
```diff
@@ -46,9 +42,8 @@ def self.create_from_csv_data(attrs)
     measurement_attrs[:name] = attrs.fetch(:measurement)
     measurement_attrs[:processed_file_id] = attrs.fetch(:processed_file_id)
     measurement_attrs[:organization_id] = attrs.fetch(:organization_id)
-    measurement_attrs[:tank_id] = tank.id if tank
-    measurement_attrs[:animal_id] = animal.id if animal
-    measurement_attrs[:family_id] = family.id if family
+    measurement_attrs[:subject] = tank.id
```
mdworken, line 45: This should not be hardcode to be a tank.

spec/controllers/file_uploads_controller_spec.rb
```diff
@@ -11,6 +11,7 @@
 
   describe '#upload' do
     it 'should upload multiple files successfully' do
+      skip
```
mdworken, line 14: @benreyn @CraigJZ why were these tests skipped?

PR #383, opened 2020-09-23
[344] Allows deletion of users

### LINKED ISSUE #344: As an admin, I can delete a user for an organization
- [ ] The UsersController `destroy` action is added
- [ ] A button/link is added to the Users#index page created in #340
- [ ] This action should ask for confirmation before this destructive action is taken
- [ ] The current_user should not be allowed to perform this action on their own record
- [ ] A feature test is added that seeds a user record in the database, loads the Users#index page and clicks to delete the record, then asserts the record is no longer on the page

This task is dependent on #339 and #340

viniciusgama:
# Checklist:

Resolves #344 

### Type of change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ schema_migrate: database_started
 schema_status:
 	docker-compose run --rm pending_migrations
 
-database_seeds:
+database_seeds: schema_migrate
 	docker-compose run --rm seeds
 
 build:
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -3,7 +3,7 @@ class UsersController < ApplicationController
   before_action :set_user, only: [:show, :destroy]
 
   def index
-    @users = current_organization.users.reject { |user| user == current_user }
+    @users = current_organization.users
   end
 
   def show; end
@@ -26,9 +26,14 @@ class UsersController < ApplicationController
   def update; end
 
   def destroy
-    @user.destroy
+    notice_message = 'User cannot delete itself.'
 
-    redirect_to users_url, notice: 'User was successfully destroyed.'
+    unless @user == current_user
+      @user.destroy
+      notice_message = 'User was successfully destroyed.'
+    end
+
+    redirect_to users_url, notice: notice_message
   end
 
   private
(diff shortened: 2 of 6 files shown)


CraigJZ (commented):
Thanks for making that change to the Makefile!!

A thought I have on the change to the index action is that it prevents an admin user from potentially reaching their own edit view, should they want to make a change to themselves.

An alternative approach might be a restriction in the destroy action that prevents @user.destroy if @user == current_user. What are your thoughts?

I also think we could merge this as is and spit handling an admin editing themself into a different story.

viniciusgama:
> Thanks for making that change to the Makefile!!
> 
> A thought I have on the change to the index action is that it prevents an admin user from potentially reaching their own edit view, should they want to make a change to themselves.
> 
> An alternative approach might be a restriction in the destroy action that prevents `@user.destroy` if `@user == current_user`. What are your thoughts?
> 
> I also think we could merge this as is and split handling an admin editing themself into a different story.

Yeah, you are right, I just saw that the edit story would have to allow a user to update itself. I actually started by adding that check in the destroy action but thought it would be easier to hide the user.
I think I'll go with your suggestion. Another option would be to disable or hide the delete button when `user == current_user`

CraigJZ:
> > Thanks for making that change to the Makefile!!
> > A thought I have on the change to the index action is that it prevents an admin user from potentially reaching their own edit view, should they want to make a change to themselves.
> > An alternative approach might be a restriction in the destroy action that prevents `@user.destroy` if `@user == current_user`. What are your thoughts?
> > I also think we could merge this as is and split handling an admin editing themself into a different story.
> 
> Yeah, you are right, I just saw that the edit story would have to allow a user to update itself. I actually started by adding that check in the destroy action but thought it would be easier to hide the user.
> I think I'll go with your suggestion. Another option would be to disable or hide the delete button when `user == current_user`

Thanks! I'd be fine with either of those options.

viniciusgama:
> Thanks! I'd be fine with either of those options.

I went with both options, I thought they would complement each other nicely. You can see the changes in there 8a687b39959997b7c215b6fc9f41afb82909575d @CraigJZ

CraigJZ (approved):
Thanks!! Looks great!.

PR #394, opened 2020-09-23
[343] Implements user edit functionality

### LINKED ISSUE #343: As an admin, I can update a user for an organization
- [ ] The UsersController `edit` and `update` actions are added
- [ ] A new view is added that renders a form, pre-populated with the data for the existing User record, for the admin to edit. Ideally this could re-use the markup for the form created in #341 
- [ ] This form can be submitted to update the user record in the database.
- [ ] A feature test is added that seeds the database with a user record, loads this page, makes an edit to that record, and submits the form

Some other considerations are that users should be managing their own passwords. One approach to this would be to give the current_user the ability to edit their own user record. This can be addressed here or split into a separate issue.

This task is dependent on #339

viniciusgama:
# Checklist:

Resolves #343 

### Description
As an admin I want to update a user for an organization

### Type of change

### How Has This Been Tested?
Feature tests implemented and also some basic exploratory tests.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,6 +1,6 @@
 class UsersController < ApplicationController
-  before_action :admin?, only: [:index, :show, :new, :create, :edit, :update, :destroy]
-  before_action :set_user, only: [:show, :destroy]
+  before_action :admin?, only: %i[index show new create edit update destroy]
+  before_action :set_user, only: %i[show edit update destroy]
 
   def index
     @users = current_organization.users
@@ -23,7 +23,13 @@ class UsersController < ApplicationController
 
   def edit; end
 
-  def update; end
+  def update
+    if @user.update(user_params)
+      redirect_to @user, notice: 'User was successfully updated.'
+    else
+      render :edit
+    end
+  end
 
   def destroy
     notice_message = 'User cannot delete itself.'
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,6 +1,6 @@
 class UsersController < ApplicationController
-  before_action :admin?, only: %i[index show new create edit update destroy]
-  before_action :set_user, only: %i[show edit update destroy]
+  before_action :admin?, only: [:index, :show, :new, :create, :edit, :update, :destroy]
+  before_action :set_user, only: [:show, :edit, :update, :destroy]
 
   def index
     @users = current_organization.users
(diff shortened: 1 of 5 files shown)


viniciusgama:
@CraigJZ @benreyn for this I thought that the regular user being able to update his own password could go in a different issue. We could implement something that doesn't expect the user to go through the users page in order to do that so we can avoid some complexity in the code. A page that allows only the password to be edited could be a lot simpler. Either way, I think this can be discussed in another issue.

If you feel there is no need for discussion and want me to implement in this PR, I'll be more than happy to do so.

CraigJZ:
> @CraigJZ @benreyn for this I thought that the regular user being able to update his own password could go in a different issue. We could implement something that doesn't expect the user to go through the users page in order to do that so we can avoid some complexity in the code. A page that allows only the password to be edited could be a lot simpler. Either way, I think this can be discussed in another issue.
> 
> If you feel there is no need for discussion and want me to implement in this PR, I'll be more than happy to do so.



> @CraigJZ @benreyn for this I thought that the regular user being able to update his own password could go in a different issue. We could implement something that doesn't expect the user to go through the users page in order to do that so we can avoid some complexity in the code. A page that allows only the password to be edited could be a lot simpler. Either way, I think this can be discussed in another issue.
> 
> If you feel there is no need for discussion and want me to implement in this PR, I'll be more than happy to do so.

Moving that piece into a different issue sounds good to me. I like the path that you proposed for it. The one piece that I would add to the discussion would be an extra layer safety around the allowed params if we were still giving a user access to the update action for their password change. One approach would be to only pass role into the allow params if current_user is an admin but I'd welcome other thoughts around that.

CraigJZ (approved):

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,6 @@
 class UsersController < ApplicationController
-  before_action :admin?, only: [:index, :show, :new, :create, :edit, :update, :destroy]
-  before_action :set_user, only: [:show, :destroy]
+  before_action :admin?, only: %i[index show new create edit update destroy]
+  before_action :set_user, only: %i[show edit update destroy]
```
wadewinningham, line 3: I think this is more a function of our editors' and rubocop setup than abalone's. Works either way, just not sure if we're injecting our code-style into their project.

wadewinningham (approved):

app/controllers/users_controller.rb
viniciusgama: you're right @wadewinningham . I think I'll revert this to keep it consistent with the rest of the codebase.

haydenrou (approved):

Issue #390, opened 2020-09-23
Add version tracking to Cohort


NiquiCara:
There are a number of things we don't know yet about what we are going to need to track over-time in this application. In the future we may have to introduce new concepts or modify the way we have modeled some concepts. In order to set us up for the best possible chance of having data to migrate into these new concepts, we should track changes accross our core models. https://github.com/paper-trail-gem/paper_trail and https://github.com/collectiveidea/audited are the most commonly used gems for this, so lets use one of them

PR #396, opened 2020-09-23
Add versioning with paper trail

### LINKED ISSUE #388: Add version tracking to Animal
There are a number of things we don't know yet about what we are going to need to track over-time in this application. In the future we may have to introduce new concepts or modify the way we have modeled some concepts. In order to set us up for the best possible chance of having data to migrate into these new concepts, we should track changes accross our core models. https://github.com/paper-trail-gem/paper_trail and https://github.com/collectiveidea/audited are the most commonly used gems for this, so lets use one of them

haydenrou:
Added to Animal, Family, Facility and Tanks

**Description**
This PR just adds the version tracking for the above models. It does not present these details anywhere.

# Checklist:

-->

Resolves #388 #389 #390 

### Type of change

work as expected)
* Documentation update


app/models/family.rb
```diff
@@ -1,19 +1,21 @@
 class Family < ApplicationRecord
+  has_paper_trail
```
haydenrou, line 2: @benreyn I added paper trail here too as it was fairly straightforward. Are you happy with this, or would you prefer it to just be on the 3 mentioned? Even further, if you wanted it on any other models it would be a very quick commit?

app/models/family.rb
```diff
@@ -1,19 +1,21 @@
...
+
   include OrganizationScope
 
-  belongs_to :male, class_name: 'Animal'
-  belongs_to :female, class_name: 'Animal'
+  belongs_to :male, class_name: 'Animal', optional: true
```
haydenrou, line 6: I've made these optional per the discussion this morning whereby you mentioned that a family could potentially not have a determined male or female in the beginning @benreyn is that okay?

app/models/family.rb
```diff
@@ -1,19 +1,21 @@
...
+  belongs_to :female, class_name: 'Animal', optional: true
 
   has_many :measurements
 
   belongs_to :tank, required: false
 
   def female_pii_tag
-    female.pii_tag
+    female&.pii_tag
```
haydenrou, line 14: Due to above comment.

db/schema.rb
```diff
@@ -10,15 +10,15 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_09_23_153522) do
```
haydenrou, line 13: There are some oddities in the schema here, which will be resolved when we merge #379

app/models/family.rb
benreyn: This is great, I think I intended to have it here as well but just missed making the issue.

benreyn (approved):
I think this is in good shape, but lets merge #379 first so we can clean up the diff on this commit

app/models/family.rb
benreyn: Yup. Good call. Under normal circumstances I probably would have done this in a separate PR, but fine to leave in here.

app/models/family.rb
viniciusgama: have you considered a delegate here @haydenrou ?

```suggestion
    delegate :pii_tag, to: :female, prefix: true, allow_nil: true
```

haydenrou: Good idea, I was initially just fixing the broken page but i'll make this refactor too

benreyn:
Looks like sadly there were some pretty big conflicts with this PR and the tank and family renames. 😞

PR #401, opened 2020-09-24
Allow user to change its own password

### LINKED ISSUE #395: As a non-admin user, I can update my own password.
We want to allow non-admin users to update their own password from within the application. They should be able to do this without being able to update their other attributes, particularly `:role`.

See discussion on PR #394

Acceptance Criteria:

- [ ] Add interface for a non-admin user to update their password
- [ ] Restrict the user from updating other attributes
- [ ] A unit specs and/or feature specs to test both of these behaviors

viniciusgama:
# Checklist:

Resolves #395 

### Description
Guide questions:
  - What motivated this change (if not already described in an issue)? Currently the admin user can update its own password through the users management page. We also need to allow a regular user to do the same.
  - What alternative solutions did you consider? Allow the regular user to view the users admin page
  - What are the tradeoffs for your solution? Less complexity on dealing with different levels of access in the same page.
   
None

### Type of change

### How Has This Been Tested?

Feature tests implemented plus some basic manual exploratory tests.

### Screenshots

Screen Shot 2020-09-24 at 10 57 34 AM
Screen Shot 2020-09-24 at 10 58 34 AM

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -57,10 +57,10 @@ class UsersController < ApplicationController
   private
 
   def user_params
+    allowed_params = [:email, :password]
+    allowed_params << :role if current_user.admin?
     params.require(:user).permit(
-      :email,
-      :password,
-      :role
+      allowed_params
     ).merge(organization_id: current_organization.id)
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,7 +1,6 @@
 class UsersController < ApplicationController
   before_action :admin?, only: [:index, :show, :new, :create, :edit, :update, :destroy]
-  before_action :set_user, only: [:show, :edit, :update, :destroy, :edit_password, :update_password]
-  before_action :current_user?, only: [:edit_password, :update_password]
+  before_action :set_user, only: [:show, :edit, :update, :destroy]
 
   def index
     @users = current_organization.users
@@ -43,17 +42,6 @@ class UsersController < ApplicationController
     redirect_to users_url, notice: notice_message
   end
 
-  def edit_password; end
-
-  def update_password
-    if @user.update(user_params)
-      sign_in(@user, bypass: true)
-      redirect_to edit_password_users_url, notice: 'Password was successfully updated.'
-    else
-      render :edit_password
-    end
-  end
-
   private
 
   def user_params
@@ -70,9 +58,5 @@ class UsersController < ApplicationController
 
   def admin?
     redirect_to root_path, alert: "Not authorized" unless current_user.admin?
-  end
-
-  def current_user?
-    redirect_to root_path, alert: "Not authorized" unless current_user == @user
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -45,10 +45,10 @@ class UsersController < ApplicationController
   private
 
   def user_params
-    allowed_params = [:email, :password]
-    allowed_params << :role if current_user.admin?
     params.require(:user).permit(
-      allowed_params
+      :email,
+      :password,
+      :role
     ).merge(organization_id: current_organization.id)
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,6 +1,7 @@
 class UsersController < ApplicationController
   before_action :admin?, only: [:index, :show, :new, :create, :edit, :update, :destroy]
-  before_action :set_user, only: [:show, :edit, :update, :destroy]
+  before_action :set_user, only: [:show, :edit, :update, :destroy, :edit_password, :update_password]
+  before_action :current_user?, only: [:edit_password, :update_password]
 
   def index
     @users = current_organization.users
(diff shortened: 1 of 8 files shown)


CraigJZ (commented):
Thanks!! I like this approach. :)

What are your thoughts on adding additional protection for permitting updates to `:role`? Maybe something like:

```
def user_params
  allowed_params = [:email, :password]
  allowed_params << :role if current_user.admin?
  params.require(:user).permit(
    allowed_params
  ).merge(organization_id: current_organization.id)
end
```

viniciusgama:
> Thanks!! I like this approach. :)
> 
> What are your thoughts on adding additional protection for permitting updates to `:role`? Maybe something like:
> 
> ```
> def user_params
>   allowed_params = [:email, :password]
>   allowed_params << :role if current_user.admin?
>   params.require(:user).permit(
>     allowed_params
>   ).merge(organization_id: current_organization.id)
> end
> ```

Great idea @CraigJZ ... here you go ebd3480052e185f0e40361ff23943562b64963c4

app/controllers/users_controller.rb
```diff
@@ -42,13 +43,24 @@ def destroy
...
+
+  def update_password
+    if @user.update(user_params)
+      sign_in(@user, bypass: true)
+      redirect_to edit_password_users_url, notice: 'Password was successfully updated.'
+    else
+      render :edit_password
+    end
+  end
+
```
benreyn, line 56: I tend to prefer to keep my controllers and routes as RESTful as possible. i.e Only having controllers with the actions that are generated for free from `resources` or `resource` in the routes file. How would you feel about pulling this out into a `PasswordsController#edit` and `PasswordsController#update` action?

app/controllers/users_controller.rb
```diff
@@ -42,13 +43,24 @@ def destroy
...
   private
 
   def user_params
+    allowed_params = [:email, :password]
+    allowed_params << :role if current_user.admin?
     params.require(:user).permit(
-      :email,
-      :password,
-      :role
+      allowed_params
```
benreyn, line 63: If we use a net new controller for this, we dont need to worry about the conditional modification of the params allowlist.

CraigJZ (approved):
Awesome!! Thanks!!

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,7 @@
 class UsersController < ApplicationController
   before_action :admin?, only: [:index, :show, :new, :create, :edit, :update, :destroy]
-  before_action :set_user, only: [:show, :edit, :update, :destroy]
+  before_action :set_user, only: [:show, :edit, :update, :destroy, :edit_password, :update_password]
+  before_action :current_user?, only: [:edit_password, :update_password]
```
benreyn, line 4: I tend to start to get worried when controllers state collection alot of `before_action` filters. It starts to make it difficult to folow the code path. "ok, what all method run before this controller action?", This gets even worse when we start having some before filters use `except:` and other use `only:`. If we pull  this into its own controller we could clean this chunk of code up some.

benreyn (commented):
Awesome! Thanks  @viniciusgama. I had some thoughts on how we could clean this up a bit by keeping the routes more REST-ful. Let me know what you think.

CraigJZ:
@benreyn Thanks for adding those thoughts. I do prefer that over the path that I'd proposed. @viniciusgama Sorry for the confusion!!

viniciusgama:
thanks for the feedback @benreyn . I'm all for moving the password actions to its own controller, even though it won't be so restful as the password is not a resource/model and instead just an attribute in the user. In that case we would have the user model being manipulated in more than one controller. But I feel we have more pros than cons in this approach anyways so I'll go ahead and move forward with it.

viniciusgama:
@CraigJZ @benreyn moved everything to a `PasswordsController`. Let me know what you think.

config/routes.rb
```diff
@@ -1,6 +1,7 @@
 Rails.application.routes.draw do
   devise_for :users, path_prefix: 'app'
   resources :users
+  resources :passwords
```
benreyn, line 4: Can we restrict this to.. 
```suggestion
  resources :passwords, only: [:edit, :update]
```
So we arent adding routes for controller actions that dont exist.

benreyn (approved):
> even though it won’t be so restful as the password is not a resource/model and instead just an attribute in the user.

Just so we’re on the same page about the approach here. I wanted to share one of my favorite blog posts on this topic...
https://thoughtbot.com/blog/let-it-rest

I think as developer’s we have extra information that the user doesn’t, and sometimes we think of a “resource” as only something that maps to one specific model. But to me (personally) we designing an API, I think of the “resources” as concepts, and since I am in the developer seat, I get to decide what constitutes a resource. 

Lets say a requirement comes down the pipeline that "As admin, I can click a button that lets turns a user into an admin. We could add to this controller again, or we could add the “conceptual” resource of “AdminUsers” and create a `AdminUsersController` with a `create` action. To me, this makes it easier to reason about all the things that your application does just by looking at the routes file.

I hope that clarifies my position some. :smile: If you feel differently about this Id love to hear your thoughts. This is one of the things that I love about open source as it gives us as developers a chance to learn from and expose ourselves to new ideas from others that we would normally never get a chance to work with. :heart:

Issue #378, opened 2020-09-23
Rename Family model to Cohort


benreyn:
We've found that the concept of a `Family` is maybe a bit too specific for our application, and we want to rename it to `Cohort`

PR #402, opened 2020-09-24
Rename families and tanks

### LINKED ISSUE #377: Rename Tank model to Enclosure
We've found that the concept of a `Tank` is maybe a bit too specific for our application, and we want to rename it to `Enclosure`

miniyakkos:
Resolves #377 #378 

### Description
Renamed Tank to Enclosure and Family to Cohort

### Type of change
* Refactoring
work as expected)

### How Has This Been Tested?
Ran entire test suite and did some manual testing through UI


benreyn (approved):
I think we are in a place where we can trust the tests that we didnt miss any renames here. Do you agree @CraigJZ ?

PR #407, opened 2020-09-24
Add shl to animals

### LINKED ISSUE #387: Add SHL_number to Animal
Animals from each spawning date and from each wild collection have a unique case number created by California's state Shellfish Health Laboratory (SHL).  Sometimes animals from a single spawning date have more than one SHL number.

I think there are a couple pieces to discuss here. The SHL_number terminology is probably unique to shellfish. Should we use a more generic column name here? Since animals can have more than 1 shl_number, we may want to consider having this be an array column.

haydenrou:
For this issue we decided to have a joins table in order to make reporting in the future easier. The thought process was that serialized columns can be difficult to do certain reporting on, especially if you wanted to check for instance which animals had which codes.

# Checklist:

-->

Resolves #387 

### Type of change

work as expected)
* Documentation update


haydenrou:
@benreyn @CraigJZ we have added the basic functionality for adding SHL numbers to animals, but we had a question regarding CSV import. Firstly, do we want to upload SHL numbers via CSV? Would it be a separate import or part of the animals one?
Currently the animal csv import is for columns on the animal and we were wondering if we would have to do some escaping for them to do a comma separated list within the CSV of different SHL numbers. 

Is this Okay for a first pass, should we expand on it or create another task in the backlog to do this?

app/controllers/animals_controller.rb
```diff
@@ -12,35 +12,36 @@ def new
...
-      if @animal.save
-        format.html { redirect_to @animal, notice: 'Animal was successfully created.' }
-        format.json { render :show, status: :created, location: @animal }
-      else
-        format.html { render :new }
-        format.json { render json: @animal.errors, status: :unprocessable_entity }
-      end
+    params[:animal][:shl_numbers_codes].split(",").each do |code|
+      @animal.shl_numbers.find_or_initialize_by(code: code)
+    end
```
benreyn, line 17: thoughts on extracting this param transformation logic into a private method on this controller?

app/controllers/animals_controller.rb
```diff
@@ -12,35 +12,36 @@ def new
   def create
     @animal = Animal.new(animal_params)
 
-    respond_to do |format|
-      if @animal.save
-        format.html { redirect_to @animal, notice: 'Animal was successfully created.' }
-        format.json { render :show, status: :created, location: @animal }
-      else
-        format.html { render :new }
-        format.json { render json: @animal.errors, status: :unprocessable_entity }
```
benreyn, line 21: Thanks for going ahead and clearing out the unused json endpoints here 👍

db/schema.rb
```diff
@@ -211,6 +226,17 @@
...
+    t.string "item_type", null: false
+    t.bigint "item_id", null: false
+    t.string "event", null: false
+    t.string "whodunnit"
+    t.text "object"
+    t.datetime "created_at"
+    t.text "object_changes"
+    t.index ["item_type", "item_id"], name: "index_versions_on_item_type_and_item_id"
+  end
+
```
benreyn, line 239: I think these are some schema changes we dont want in this diff.

spec/models/shl_number_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe ShlNumber, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
benreyn, line 4: ✂️

spec/models/animals_shl_number_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe AnimalsShlNumber, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
benreyn, line 4: ✂️

benreyn (commented):
Looks great! Thanks for tackling this so quickly @haydenrou. Had a few comments. Let me know what you think.

At a higher level, what are your thoughts on adding to the existing feature tests for creating an animal and updating an animal so that we can ensure we dont break this functionality in the future?

benreyn:
> Is this Okay for a first pass, should we expand on it or create another task in the backlog to do this?


I think this is fine for a first pass, if we need  to expand the csv import functionality to include SHL numbers in the future, then we can do so then. No need to add more functionality than we will know we need.

db/schema.rb
haydenrou: This should be sorted when #396 is merged

todtb (approved):
lgtm!

PR #409, opened 2020-09-24
[386] Set up Locations

### LINKED ISSUE #386: Introduce missing concept: Location
In meeting with our stakeholders we realized that there was a concept that we were missing that lie between a Facility and an Enclosure(Tank). 

Facilities may (or may not) have_many locations within that given facility where Enclosures, Cohort's, and Animals are kept. We will want our users to know _where_ in a facility a given Cohort or Echlosure is located. 

We should likely break up the relationship of Enclosure to Facility, and instead let an Enclosure belong to a Facility _through_ a Location.

bkrigel:
# Checklist:

Resolves #386

### Description
Introduces the `Location` model to the application, associates Location to Facility and Enclosure, write related feature specs

### Type of change

### How Has This Been Tested?
Feature specs to test the index, show, new, update, and destroy actions

### Screenshots
Screen Shot 2020-09-24 at 1 52 01 PM

### DIFF
--- a/spec/fixtures/enclosures.csv
+++ b/spec/fixtures/enclosures.csv
@@ -1,6 +1,6 @@
-facility_code,name
-AOP,Enclosure Number 1
-AOP,Enclosure Number 2
-AOP,Enclosure Number 2
-QQQ,Enclosure Number 1
-,
+facility_code,location_name,name
+AOP,SeaLab2020,Enclosure Number 1
+AOP,Atlantis,Enclosure Number 2
+AOP,SeaWorld,Enclosure Number 2
+QQQ,OceanPlace,Enclosure Number 1
+,,
--- a/spec/models/location_spec.rb
+++ b/spec/models/location_spec.rb
@@ -1,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Location, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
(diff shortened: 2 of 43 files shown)


spec/models/location_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Location, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
benreyn, line 5: Are there any tests you think would be valuable to add here? if not, should we just delete this file until we do?

spec/fixtures/enclosures.csv
```diff
@@ -1,6 +1,6 @@
...
-AOP,Enclosure Number 2
-AOP,Enclosure Number 2
-QQQ,Enclosure Number 1
-,
\ No newline at end of file
+facility_code,location_name,name
+AOP,SeaLab2020,Enclosure Number 1
+AOP,Atlantis,Enclosure Number 2
+AOP,SeaWorld,Enclosure Number 2
+QQQ,OceanPlace,Enclosure Number 1
```
benreyn, line 5: 🤣

benreyn (approved):
There are ALOT of lines of code here so I only really skimmed this PR, but everything seems to be in order. @CraigJZ, would you mind doing a second pass?

spec/models/location_spec.rb
smkopp92: Good catch, thanks Ben!

PR #410, opened 2020-09-24
Reworked animal tags to be in a single string column

### LINKED ISSUE #397: Tags: Rework how Animals are identified within an organization
Currently Animals have a few different attributes that are used as identifiers. We have both `tag_id` and `pii_tag`, neither of which _really_ meets our needs. After talking with our stakeholders, we have learned that a `tag` on an animal is _really_ a combination of the color of a tag and a three digit number on that tag. So an individual animal might be identified by "Green 127" or "Yellow 341". These tags are usually written in lab notes or recorded in excel files as shorthand, `G127` or `Y341` from the examples here.

Tags are also, not always unique to animal, but they ARE always unique to an animal within a Cohort. There may be two animals identified as `G127`, but never two animals that belong to the same cohort. As such, we _believe_ that an animal can be uniquely identified by a combination of their tag and the cohort they belong to.

haydenrou:
Co-authored-by: Vinicius Gama <vinicius.sgama@gmail.com>

# Checklist:

-->

Resolves #397 

### Type of change

work as expected)
* Documentation update

### DIFF
--- a/spec/factories/animals.rb
+++ b/spec/factories/animals.rb
@@ -3,7 +3,7 @@ FactoryBot.define do
     collection_year { 1 }
     date_time_collected { "2020-04-04 18:52:09" }
     collection_position { "MyString" }
-    sequence(:tag) { |n| "G#{n}" }
+    sequence(:tag) { |n| "G#{sprintf('%03d', n)}" }
     sex { Animal.sexes.keys.sample }
     organization
     cohort
@@ -12,7 +12,7 @@ FactoryBot.define do
     collection_year { 1 }
     date_time_collected { "2020-04-04 18:52:09" }
     collection_position { "MyString" }
-    sequence(:tag) { |n| "Y#{n}" }
+    sequence(:tag) { |n| "Y#{sprintf('%03d', n)}" }
     sex { 'male' }
     organization
   end
@@ -20,7 +20,7 @@ FactoryBot.define do
     collection_year { 1 }
     date_time_collected { "2020-04-04 18:52:09" }
     collection_position { "MyString" }
-    sequence(:tag) { |n| "R#{n}" }
+    sequence(:tag) { |n| "R#{sprintf('%03d', n)}" }
     sex { 'female' }
     organization
   end
--- a/spec/factories/animals.rb
+++ b/spec/factories/animals.rb
@@ -3,7 +3,7 @@ FactoryBot.define do
     collection_year { 1 }
     date_time_collected { "2020-04-04 18:52:09" }
     collection_position { "MyString" }
-    sequence(:tag) { |n| "G#{sprintf('%03d', n)}" }
+    sequence(:tag) { |n| "G#{format('%<number>03d', number: n)}" }
     sex { Animal.sexes.keys.sample }
     organization
     cohort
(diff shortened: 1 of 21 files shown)


spec/factories/animals.rb
```diff
@@ -13,17 +12,15 @@
     collection_year { 1 }
     date_time_collected { "2020-04-04 18:52:09" }
     collection_position { "MyString" }
-    pii_tag { 1 }
-    tag_id { 1 }
+    sequence(:tag) { |n| "Y#{n}" }
```
benreyn, line 15: So that factory data maps more closely to what we'll see in the real world, how would you feel about left-padding the `n` here? 
```suggestion
    sequence(:tag) { |n| "Y#{sprintf('%03d', 1)}" }
```

spec/factories/animals.rb
```diff
@@ -13,17 +12,15 @@
...
     sex { 'male' }
     organization
   end
   factory :female, class: Animal do
     collection_year { 1 }
     date_time_collected { "2020-04-04 18:52:09" }
     collection_position { "MyString" }
-    pii_tag { 1 }
-    tag_id { 1 }
+    sequence(:tag) { |n| "R#{n}" }
```
benreyn, line 23: Same suggestion here about left-padding the tag... 
```suggestion
    sequence(:tag) { |n| "R#{sprintf('%03d', 1)}" }
```

spec/features/animals/create_spec.rb
```diff
@@ -13,7 +13,7 @@
     visit new_animal_path
 
     within('form') do
-      fill_in 'animal_pii_tag', with: "An Animal"
+      fill_in 'animal_tag', with: "An Animal"
```
benreyn, line 16: So that our tests more closely match real world usage, thoughts on this change?
```suggestion
      fill_in 'animal_tag', with: "G127"
```

spec/fixtures/animals.csv
```diff
@@ -1,4 +1,4 @@
-collection_year,date_time_collected,collection_position,pii_tag,tag_id,sex,unknown_column
-2020,2020-09-22T08:16:24-04:00,sample,1,0,female,unknown_data
-2020,2020-09-22T08:16:24-04:00,sample,2,0,female,unknown_data
-2020,2020-09-22T08:16:24-04:00,sample,2,0,female,unknown_data
\ No newline at end of file
+collection_year,date_time_collected,collection_position,tag,sex,unknown_column
+2020,2020-09-22T08:16:24-04:00,sample,G1,female,unknown_data
+2020,2020-09-22T08:16:24-04:00,sample,Y2,female,unknown_data
+2020,2020-09-22T08:16:24-04:00,sample,Y2,female,unknown_data
```
benreyn, line 4: Similar suggestion here about the code / data more closely resembling reality
```suggestion
2020,2020-09-22T08:16:24-04:00,sample,G001,female,unknown_data
2020,2020-09-22T08:16:24-04:00,sample,Y002,female,unknown_data
2020,2020-09-22T08:16:24-04:00,sample,Y2,female,unknown_data
```

spec/features/animals/upload_spec.rb
```diff
@@ -22,7 +22,7 @@
     within('tbody') { expect(page).to have_xpath('.//tr', count: 2) }
 
     # Verify no duplicates and assigned to correct org
-    animals = Animal.where(pii_tag: 2)
+    animals = Animal.where(tag: "Y2")
```
benreyn, line 25: If you take my suggestion about the factory above, you'll probably have a test to fix here.

benreyn (approved):
Awesome! Thanks @haydenrou. I had some suggestions about how we could help make the test data more closely  resemble real data, but they are just suggestions. Let me know what you think

I think I ultimately like the, "Just use a string column" approach. It makes this more flexible for the future, while giving us the data we would need to break it up in the future if we want to. 🥇

PR #412, opened 2020-09-25
Add templates, ui for measurement csv uploads

### LINKED ISSUE #307: CSV upload for measurement
Waiting on sample data files from stakeholders

- [x] Generate template based on user input 
#412 

- [ ] Process uploaded file

todtb:
Adds two templates for measurement csv uploads along with an interface
that can be used to select and download a sample csv

Hey @benreyn and @CraigJZ. Looking to get some feedback on whether or not this was the approach you had in mind for  #307. Obviously we can adjust the templates as needed. But also want to make sure the selections in the UI were what you had in mind, or if there was something I'm missing here. Let me know your thoughts!

Partially resolves  #307 

### Description

First step for #307. This provides csv templates for the upload of measurements. It includes headers and a few rows of sample data.

### Type of change

### How Has This Been Tested?
Manually, specs will follow with the next task which is the uploading and parsing of completed csv templates.

### DIFF
--- a/app/views/measurements/new.html.erb
+++ b/app/views/measurements/new.html.erb
@@ -1,0 +1,17 @@
+<section class="section facilities__form">
+  <div class="container facilities__form__container">
+    <h1 class="title">New Measurements</h1>
+    <div class="facilities__form__container__body">
+      <div class="field">
+        <%= select_tag(:data_import_type, options_for_select([
+             ['I am taking measurements of animals', :animal_measurements],
+             ['I want to measure length and gonad score', :length_gonad_score]
+        ])) %>
+      </div>
+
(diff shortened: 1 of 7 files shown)


haydenrou (approved):

app/views/measurements/new.html.erb
```diff
@@ -0,0 +1,17 @@
+<section class="section facilities__form">
```
haydenrou, line 1: Do these need to be changed to Tailwind actually?

todtb: Sounds good. Thought it would make sense to be a new branch to go with our other style PRs

Issue #391, opened 2020-09-23
As a user, I can view a report that represents the total number of live animals at a specific time broken out by Cohort


NiquiCara:
Using the solution from either #335 or #337 implement this report required by stakeholder.

wadewinningham:
Done with #337

wadewinningham:
https://github.com/rubyforgood/abalone/pull/414

PR #414, opened 2020-09-25
Add ReportsKit with Total Animals by Cohort report

### LINKED ISSUE #337: Install ReportsKit and configure it for organization level scoped reports.
While with #335 we are looking to setup an environment where we can run reports and create dashboards right inside the app with raw SQL queries, we also will need to create reports that are more of a "first class concept" in the application. I believe that [ReportsKit](https://www.reportskit.co/) will allow us to develop these Reports and Dashboards quickly and allow our users to filter and analyze these reports by the metrics they will want (especially those that don't know SQL, assuming we can get organization level scoped access working for Blazer).

We will need a users reports to be scoped to the current logged in users organization. In my initial research, https://github.com/tombenner/reports_kit/issues/7 should give an accurate description of how to accomplish these kinds of "scoped" reports using ReportsKit.

wadewinningham:
Adds ReportsKit gem to the app along with adding a Total Animals by Cohort chart that's filterable by cohort and date range (based on created_at)

# Checklist:

Resolves #337 and #391 

### Description

The [ReportsKit gem](https://www.reportskit.co) is added here to provide chart support for the application.

### Type of change

### How Has This Been Tested?

Locally, I generated a sampling of animal data belonging to various cohorts to use to test with.

### Screenshots

Screen Shot 2020-09-25 at 9 55 23 AM


app/models/animal.rb
```diff
@@ -3,6 +3,13 @@ class Animal < ApplicationRecord
 
   include OrganizationScope
   include CsvExportable
+  include ReportsKit::Model
+
+  reports_kit do
```
haydenrou, line 8: Could we potentially move this up to the `OrganizationScope` concern, or is it likely not going to be used on many?

haydenrou (approved):

app/models/animal.rb
wadewinningham: I tried that. The problem is that a model only accepts a single `reports_kit do` block. I put it here because I anticipate needing to add some more detail here that other models wouldn't share.

viniciusgama (approved):
Looks good. Just a couple of rubocop issues

miniyakkos (approved):

darrendc (approved):

PR #439, opened 2020-10-01
[#436] Add Facility model validation

### LINKED ISSUE #436: Add model level validations to Facility
Add model level validations to Facility

- [ ] name: presence
- [ ] code: presence and uniqueness (scoped to organization)

@CraigJZ Do these sound reasonable to you?

nsiregar:
Resolves #436  

### Description

Add model level validation to `Facility` which do
- validates the presence of the name
- validates the presence of code
- validates the uniqueness of code, scoped to organization_id
### Type of change

### How Has This Been Tested?

Tested with rspec matcher but with some issue regarding shoulda-matchers issue here https://github.com/thoughtbot/shoulda-matchers/issues/814

Instead of using `organization` object, it is scoped to `organization_id`

### DIFF
--- a/spec/models/facility_spec.rb
+++ b/spec/models/facility_spec.rb
@@ -27,9 +27,20 @@ RSpec.describe Facility, type: :model do
     it { is_expected.to have_db_column :organization_id }
 
     describe 'it only has 6 columns' do
-      it { expect(Facility.columns.count).to eq 6 }
+      it { expect(described_class.columns.count).to eq 6 }
     end
   end
 
-  include_examples 'organization presence validation'
+  describe 'validation' do
+    it 'validates presence of' do
+      is_expected.to validate_presence_of :name
+      is_expected.to validate_presence_of :code
+    end
+
+    it 'validate uniqueness of' do
+      is_expected.to validate_uniqueness_of(:code).scoped_to(:organization_id)
+    end
+  end
+
+  it_behaves_like OrganizationScope
 end
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -19,6 +19,9 @@ class Facility < ApplicationRecord
 
   has_many :locations
 
+  validates :name, presence: true
+  validates :code, presence: true, uniqueness: { scope: :organization_id }
+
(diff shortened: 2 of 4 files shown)


spec/features/facilities/index_spec.rb
```diff
@@ -29,7 +29,11 @@
   end
 
   it "should allow exporting to csv" do
-    facilities = create_list(:facility, 5, organization: user.organization)
+    facility_codes = %w[vortex anemo dendro pyro electro]
+    facilities = []
+    facility_codes.each do |code|
+      facilities << create(:facility, code: code, organization: user.organization)
+    end
```
benreyn, line 36: Instead of this, could we update the [facility factory](https://github.com/rubyforgood/abalone/blob/master/spec/factories/facility.rb) to use a [sequence](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences)? The `create_list` approach should still work in that case.

spec/models/facility_spec.rb
```diff
@@ -16,6 +16,11 @@
 RSpec.describe Facility, type: :model do
   let(:facility) { FactoryBot.create :facility }
 
+  it 'includes required modules' do
```
benreyn, line 19: These changes dont seem related to this PR. Would you mind making a separate PR for these changes that explains why these changes are needed?

benreyn (commented):
Thanks for tackling this for us @nsiregar! I left a few notes, but Ill also wait for @CraigJZ to weigh in.

spec/features/facilities/index_spec.rb
CraigJZ: Yes, thanks for taking on this issue @nsiregar. I agree, if we could use the `create_list` approach here, that would be great.

nsiregar: Sure, we can use factory sequence

spec/models/facility_spec.rb
nsiregar: Will do it

benreyn (approved):
This looks good! Thanks so much @nsiregar!

PR #440, opened 2020-10-01
[#437] Add Cohort model validation

### LINKED ISSUE #437: Add model level validations to Cohort
Add model level validations to Cohort

- [ ] name: presence and uniqueness (scoped to organization)
- [ ] Enclosure: presence

nsiregar:
Resolves #437  

### Description

Add model level validation to `Cohort` which do

- validates the presence of the name
- validates the presence of enclosure
- validates the uniqueness of name, scoped to organization_id

### Type of change

work as expected)
* Documentation update

### How Has This Been Tested?

Tested with rspec matcher but with some issue regarding shoulda-matchers issue here thoughtbot/shoulda-matchers#814

Instead of using organization object, it is scoped to organization_id

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -11,7 +11,9 @@ class Cohort < ApplicationRecord
   has_many :measurements, as: :subject
   has_many :animals
 
-  belongs_to :enclosure, required: false
+  belongs_to :enclosure, required: true
+
+  validates :name, presence: true, uniqueness: { scope: :organization_id }
 
   delegate :name, to: :enclosure, prefix: true, allow_nil: true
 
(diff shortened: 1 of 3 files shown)


benreyn:
@nsiregar Thanks for working on this! It looks like there are some test failures that adding these validations created. Could you take a look at those? We'll need TravisCI to pass before we can merge.

nsiregar:
Hi @benreyn , I was trying to understanding what is the relation between `Cohort`, `Enclosure`, and `Operation` it seems like Operation was used to update the Cohort to Enclosure. If we would like to validate the presence of `Enclosure` in `Cohort`, shouldn't it be sufficient with the `required: true` in `belongs_to: :enclosure association`?

PR #441, opened 2020-10-01
Update animal, enclosure, and measurement factories

todtb:
### Description
Recent changes to the models and associations mean that it is now possible to make more precise and complete factories for the test suite.

This change introduces or updates Animal, Enclosure and Measurement factories and minor changes to the specs to incorporate said changes.

### Type of change

### How Has This Been Tested?
As these factories are used only in the test suite currently, rspec testing should be sufficient.

### DIFF
--- a/spec/factories/animals.rb
+++ b/spec/factories/animals.rb
@@ -6,22 +6,8 @@ FactoryBot.define do
     sequence(:tag) { |n| "G#{format('%<number>03d', number: n)}" }
     sex { Animal.sexes.keys.sample }
     organization
-    cohort
-  end
-  factory :male, class: Animal do
-    collection_year { 1 }
-    date_time_collected { "2020-04-04 18:52:09" }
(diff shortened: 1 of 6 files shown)


spec/factories/animals.rb
```diff
@@ -6,22 +6,8 @@
...
-  factory :female, class: Animal do
-    collection_year { 1 }
-    date_time_collected { "2020-04-04 18:52:09" }
-    collection_position { "MyString" }
-    sequence(:tag) { |n| "R#{format('%<number>03d', number: n)}" }
-    sex { 'female' }
-    organization
+
+    factory(:male) { sex { 'male' } }
+    factory(:female) { sex { 'female' } }
```
benreyn, line 11: :heart: This cleanup!

benreyn (approved):
Thanks for tackling this @todtb! good call getting these cleaned up!

PR #442, opened 2020-10-01
Update enclosure, facility, location, op batch, and user factories

todtb:
Continuation of #441 

### Description

Recent changes to the models and associations mean that it is now possible to make more precise and complete factories for the test suite.

This change introduces or updates Facility, Location, Operation Batch, and User factories and minor changes to the specs to incorporate said changes.
### Type of change

### How Has This Been Tested?
As these factories are used only in the test suite currently, rspec testing should be sufficient.


spec/factories/users.rb
```diff
@@ -4,5 +4,8 @@
     password { "password" }
     password_confirmation { "password" }
     organization
+
+    trait(:as_admin)  { role { :admin } }
+    trait(:as_user)   { role { :user } }
```
benreyn, line 9: I think id prefer `:user` just be the default, and only define a trait for admin. And Im not sure how I feel about the `as` prefix. `as_admin` reads to me like a macro or mixin we would use for a feature test, but for the factory I think `create(:user, :admin)` reads cleaner, IMO.

spec/features/locations/index_spec.rb
```diff
@@ -8,9 +8,10 @@
   end
 
   it "Then I see a list of locations" do
-    locations = create_list(:location, 3, organization: user.organization)
+    facility = create(:facility)
+    locations = create_list(:location, 3, facility: facility, organization: user.organization)
 
-    visit facility_locations_path(locations.first.facility)
+    visit facility_locations_path(facility)
```
benreyn, line 14: :heart:

spec/factories/locations.rb
```diff
@@ -1,7 +1,16 @@
...
     facility
     organization
+
+    trait :with_enclosures do
+      transient { enclosures_count { 5 } }
+
+      after(:create) do |loc, eval|
+        create_list :enclosure, eval.enclosures_count, location: loc
+        loc.reload
+      end
```
benreyn, line 13: Its more verbose, but I think i prefer the full `location` here for clarity.

spec/factories/locations.rb
```diff
@@ -1,7 +1,16 @@
 FactoryBot.define do
   factory :location do
-    name { "MyString" }
+    sequence(:name) { |n| "Location #{n}" }
     facility
     organization
+
+    trait :with_enclosures do
+      transient { enclosures_count { 5 } }
```
benreyn, line 8: I think `1` might be better defualt here. I worry that if we have to use this trait alot we'd end up creating alot of uneeded records, slowing down the test suit as a whole.

spec/factories/facility.rb
```diff
@@ -1,8 +1,7 @@
 FactoryBot.define do
   factory :facility do
-    val = Faker::Name.unique.name
-    name { val }
-    code { val.gsub(/[aeiou|AEIOU|\s]+/, '').strip }
-    organization_id { FactoryBot.create(:organization).id }
+    name { Faker::Games::HalfLife.location.to_s }
```
benreyn, line 3: Making a note for myself to clean up usage of `Faker`. I dont think we really need it.

benreyn (commented):
Thanks for tackling this @todtb! I left some notes. Let me know what you think. Also pinging @CraigJZ

spec/factories/users.rb
todtb: How about using `:user` as a default role but keeping the trait for consistency?

spec/factories/locations.rb
todtb: I think I'll have a few changes in #445 then 😆

spec/factories/locations.rb
todtb: Makes sense

spec/factories/facility.rb
todtb: 😢  but fair enough

benreyn (approved):
Awesome! Thanks @todtb!

benreyn (approved):
Fixed some merge conflicts with #439. We may want to circle back if these create problems in the future

Issue #438, opened 2020-09-30
Add vlaidation of presence of organization to all models that require it


benreyn:
I think the fastest way to do this will be to add it to the `OrganizationScope` concern

emgreen33:
@benreyn would happily add these validations/specs if you're still looking!

PR #443, opened 2020-10-01
Add Organization presence validation to OrganizationScope and add spec coverage to relevant models

emgreen33:
Resolves https://github.com/rubyforgood/abalone/issues/438 

**Description**
1. Add Organization presence validation to OrganizationScope and add spec coverage to relevant models
In the process of writing specs, fixed the broken measurement_events factory which had incorrect associations
2. Added `.generators` to `.gitignore` as RubyMine automatically creates it

**Type of change**
Spec coverage of organization validation

**How Has This Been Tested?**
Tested with RSpec

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@
 
 .DS_Store
 .vscode
+.generators
 
 /public/packs
 /public/packs-test
--- a/app/models/concerns/organization_scope.rb
+++ b/app/models/concerns/organization_scope.rb
@@ -3,6 +3,7 @@ module OrganizationScope
 
   included do
     belongs_to :organization
+    validates :organization, presence: true
 
     scope :for_organization, ->(organization) { where(organization: organization) }
   end
--- a/spec/factories/measurement_events.rb
+++ b/spec/factories/measurement_events.rb
@@ -1,7 +1,6 @@
 FactoryBot.define do
   factory :measurement_event do
     name { "MyString" }
-    enclosure
-    organizations
+    organization
   end
 end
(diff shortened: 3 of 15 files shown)


benreyn:
This is awesome @emgreen33! Thanks for tackling this issue! We really appreciate it!

Im a little torn on the `annotate` gem stuff here. Its something Ive been meaning to bring up with the group. I tend to think its like documentation that falls out of date for too easily, and its really just a duplication of what already exists in [`schema.rb`](https://github.com/rubyforgood/abalone/blob/master/db/schema.rb), What are your thoughts @CraigJZ?

IF we _DO_ want it, I think I would prefer it get added in a separate PR, rather than conflate the git history in this PR.

emgreen33:
> This is awesome @emgreen33! Thanks for tackling this issue! We really appreciate it!
> 
> Im a little torn on the `annotate` gem stuff here. Its something Ive been meaning to bring up with the group. I tend to think its like documentation that falls out of date for too easily, and its really just a duplication of what already exists in [`schema.rb`](https://github.com/rubyforgood/abalone/blob/master/db/schema.rb), What are your thoughts @CraigJZ?
> 
> IF we _DO_ want it, I think I would prefer it get added in a separate PR, rather than conflate the git history in this PR.

Definitely agree with your thoughts here on a separate PR, and removed it from this one. If you do go that route, I believe adding `Annotate.load_tasks` to `auto_annotate_models.rake` will enable it to run automatically on each migration to prevent stale annotations, but will leave that up to you 🙂

CraigJZ:
I agree, as well, with making the annotations a separate PR. I'm ok with introducing them as long as they can be maintained in a fairly automated way.

CraigJZ (approved):

PR #451, opened 2020-10-02
Feature/update statistics on home page 434

### LINKED ISSUE #434: Update statistics on home page
The application home page currently shows a set of static statistics. We would like to change these so they pull 3 statistics from the current organization. To start, we can show the number of facilities, the number of cohorts and the number of animals. For now, let's use the same 3 statistics for all organizations, however, in the future we may want to have these differ per organization based on what is most meaningful for individual organizations.

Acceptance criteria:

- [ ] Add logic to only display the statistics when logged in.
- [ ] Add dynamically populated data for facilities, cohorts and animals based on the `current_organization`
- [ ] Add feature specs for this display

PuZZleDucK:
### Description
Pull Request for Hacktoberfest that adds dynamic information to the home page when a user is logged in and shows dummy data otherwise.
<strike>I'm not sure if the assignments in the controller should be wrapped in a conditional rather than using 'safe access' and 'or', interested in which you think would look cleaner?</strike> nevermind, RuboCop had firm opinions on it 😯 and it's now a conditional containing assignments 👍.
I also had to do some reloading of models in the tests so the relations would be initialised and work. There could be some rspec idiom or simpler way to do that?
Resolves #434 

### Type of change

### How Has This Been Tested?

Created and rand tests to verify fix. Added example data to make tests more effective. Performed manual verification with seed accounts.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -6,10 +6,6 @@ class HomeController < ApplicationController
       @facility_count = current_user.organization.facilities.count
       @cohort_count = current_user.organization.cohorts.count
       @animal_count = current_user.organization.animals.count
-    else
-      @facility_count = 3
-      @cohort_count = 7
-      @animal_count = 41
     end
   end
 
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,15 +1,9 @@
 class HomeController < ApplicationController
   skip_before_action :authenticate_user!
   def index
-    if current_user
-      @facility_count = current_user.organization.facilities.count
-      @cohort_count = current_user.organization.cohorts.count
-      @animal_count = current_user.organization.animals.count
-    else
-      @facility_count = 3
-      @cohort_count = 7
-      @animal_count = 41
-    end
+    @facility_count = current_user&.organization&.facilities&.count || 3
+    @cohort_count = current_user&.organization&.cohorts&.count || 7
+    @animal_count = current_user&.organization&.animals&.count || 41
   end
 
   def show; end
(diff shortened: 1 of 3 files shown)


todtb (approved):
Looks good, nice work. I would have gone with the safe nav myself as well 😄 

As for the reload, you can use `let!` on each of the models and then you shouldn't have to reload during the logged in run.

Regarding wording, I'm not sure if "No. of Spawn Dates" == "Cohorts". It might represent something slightly different, but maybe @benreyn or @CraigJZ can help clarify.

CraigJZ (changes requested):
@PuZZleDucK Thanks for picking up this issue! We appreciate it!!

The number of cohorts may differ from the number of spawning dates. One scenario where they might diverge is if a group of animals from an existing cohort is removed and split out into their own cohort for an experiment. We will want to change the subtitle on that card to `No. of Cohorts`. Apologies for not clarifying that change in the original issue. Thanks for checking on that.

Rather than show the fake data, we would like to, instead, not show the cards at all when a user isn't logged in.

Thanks @todtb for suggesting using `let!` to avoid the reloads.

PuZZleDucK:
Thanks @todtb, !let worked a treat 🎉.

@CraigJZ: I've updated the PR so no cards are displayed until a user logs in.

I was thinking we could use the home page cards to link to info about the project (maybe Github, Twitter and the ucdavis site) or links to the videos when a user isn't logged in? ... sing out if you want to stick anything in there, otherwise I think we're good to go :+1:

CraigJZ (approved):
Thanks @PuZZleDucK!!! This looks great.

I like your idea to use the cards to link to information about the project when a logged in user isn't present. My initial thought is that the items in the `More` (About the project, Ruby for Good and Github) navigation might work nicely there.

PR #453, opened 2020-10-04
Closes issue https://github.com/rubyforgood/abalone/issues/353.

### LINKED ISSUE #353: Remove unused JSON endpoints.
As much of the application in its current state was built with  Rails scaffolds, we have a good number of controllers that _can_ respond with JSON that do not need to. Lets clear out these unused code paths so there isnt any confusion down the road.

LittleBigProgramming:
# Checklist:

Resolves #353 

### Description
Closes issue https://github.com/rubyforgood/abalone/issues/353.

To check for the unused routes the following gem was installed locally
https://github.com/amatsuda/traceroute. The unused routes were then
checked using `rails traceroute:unused_routes`.

Reviewer should double check that no implicitly called routes have not
been removed.

### Type of change
* Removed unused endpoints, could potentially be breaking if one was implicity called

The routes that were removed are
```
Unused routes:
measurements#data_import
operations#create
operations#new
operations#edit
operations#show
operations#update
operations#update # Not sure why this was listed twice?
operations#destroy
```

### How Has This Been Tested?
As a second alternative to using `traceroute` a script was ran to double-check the unused routes output a rake task was created as advised from https://medium.com/techies-toolkit/get-unused-routes-of-your-rails-app-be602766fda9 

Incase this want's to also be tested using a rake task here is the code
```
task unused_routes: environment do
  Rails.application.eager_load!
  unused_routes = {}

  # Iterating over all non-empty routes from RouteSet
  Rails.application.routes.routes.map(&:requirements).reject(&:empty?).each do |route|
    name = route[:controller].camelcase
    next if name.start_with?("Rails")

    controller = "#{name}Controller"

    if Object.const_defined?(controller) && !controller.constantize.new.respond_to?(route[:action])
      # Get route for which associated action is not present and add it in final results
      unless Dir.glob(Rails.root.join("app", "views", name.downcase, "#{route[:action]}.*")).any?
        unused_routes[controller] = [] if unused_routes[controller].nil?
        unused_routes[controller] << route[:action]
      end
    end
  end

  puts unused_routes 
end
```

There was one route which got flagged which was left in which is `file_uploads#show_processing_csv_errors` as it was being called in `app/views/file_uploads/index.html.erb` on line:10.

If there is anything that has been missed please let me know! 🙂

### DIFF
--- a/app/controllers/measurements_controller.rb
+++ b/app/controllers/measurements_controller.rb
@@ -1,2 +1,3 @@
 class MeasurementsController < ApplicationController
+  def new; end
 end
--- a/app/controllers/measurements_controller.rb
+++ b/app/controllers/measurements_controller.rb
@@ -1,3 +1,2 @@
 class MeasurementsController < ApplicationController
-  def new; end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -22,6 +22,7 @@ Rails.application.routes.draw do
   end
 
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :measurements, only: %[new]
   post '/file_uploads', to: 'file_uploads#upload'
   resources :file_uploads, only: [:index, :new, :destroy]
   get '/file_uploads/csv_index', to: 'file_uploads#csv_index', as: 'csv_index'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -22,9 +22,6 @@ Rails.application.routes.draw do
   end
 
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
-  resources :measurements, only: %i[new]
-  post '/data_import', to: 'measurements#data_import', as: 'data_import'
-
   post '/file_uploads', to: 'file_uploads#upload'
   resources :file_uploads, only: [:index, :new, :destroy]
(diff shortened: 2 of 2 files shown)


CraigJZ (changes requested):
@LittleBigProgramming thanks for taking a look at this issue. From the CI failures, it looks like `resources :measurements, only: %i[new]` was removed by mistake. Fixing that should clear most of the spec failures.

LittleBigProgramming:
@CraigJZ Thanks for letting me know! I will take a look after work! 🙂

LittleBigProgramming:
Re-added the route and specs passed locally and on CI 🙂

CraigJZ (changes requested):
Thanks for making these updates @LittleBigProgramming. Looks like one more small change needed. The new action shouldn't be removed from the MeasurementsController.

LittleBigProgramming:
@CraigJZ Added the new action back in to the measurements_controller 👍 Let me know if there is anything else!

CraigJZ (approved):
Thanks @LittleBigProgramming!!!

CraigJZ:
Adding this note for documentation on the issue. The only remaining JSON end points are in the facilities controller.

PR #454, opened 2020-10-04
[WIP] Add Blazer gem and investigate restricting organizations to their own data.

### LINKED ISSUE #335: Install Blazer gem for Report and Dashboard generation.
https://github.com/ankane/blazer is an excellent gem that provides an out of the box solution for data exploration and generating reports and dashboard from raw SQL queries. 

👇 THIS IS A BIG SCARY VERSION OF THIS TASK 👇 

IDEALLY, we would like to give our users access to this dashboard. In order to do that, we will need to scope their access to the database, to only data that belongs to their organization. (See #336, #334, and #333). This is problematic, since Blazer sets the database access as application boot time, and we don't know the `current_user` or what `organization` they belong to until application run time. 

👇 THIS IS NOT 👇 

At the very least, as we do work for researchers working to repopulate ablone, we will need to generate reports and help them  look at trend. We can build these reports and dashboards each on a one off basis as development tasks, but having access to a tool like Blazer will allow us to prototype those reports / dashboard much faster. If the "big scary" version of this task is too large and undertaking, we should at least get this gem installed and running for admin users only. (See #93)

wadewinningham:
## Checklist:
- [ ] I have added tests that prove my fix is effective or that my feature works,
- [ ] New and existing unit tests pass locally with my changes ("bundle exec rake"),

Resolves #335 

### Description

[Issue 335](https://github.com/rubyforgood/abalone/issues/335) relates to using the [blazer gem](https://github.com/ankane/blazer) for reporting.

The problem is that there's no clean way of tying Abalone users to their own organization's data.

This PR shows a way, using Postgres row-level security, that would address that issue with relatively minimal overriding of blazer's code, paired with a bit of Postgres setup/scripting to setup appropriate users and policies per organizaiton.

This means anytime an organization is added, those security scripts would need to be re-executed. This PR provides rake tasks to do this manually, but they could be triggered after `db:migrate` in deployments if desired.

> **Disclaimer:** I am no Postgres expert. Please have someone with more skill than me review the Postgres stuff if you'd like to try this out.

Note that the security setup only pertains to tables that have an `organization_id` column. If there's a table without one, it's currently open to be used for reporting by any user.

### Type of change

* Documentation update

### How Has This Been Tested?

While I have tested this locally and it works as I think the project wants, I have not written any specs. Since the behavior is blazer itself, I'm sure we can add some integration tests. There's just some setup for such tests in that the Postgres scripts I mentioned would be required to be run.

Since I got things working, I wanted to share in case anything here in regards to maintaining the Postgres security would point the team away from using blazer. 

### Screenshots

Both screenshots are of the exact same report within blazer. The only difference is that the one with more data was user from `organization_id == 1` and the other is a user from `organization_id == 2`.

org1-user-blazer

org2-user-blazer

It's the same exact query, but with different results because of the row-level security setup within Postgres.


lib/tasks/blazer.rake
```diff
@@ -0,0 +1,154 @@
...
+      ALTER TABLE #{table} DISABLE ROW LEVEL SECURITY;
+      DROP POLICY org_policy ON #{table};
+    SQL
+  end
+
+  def policy_case_conditions
+    Organization.ids.map do |id|
+      "WHEN CURRENT_USER = 'org#{id}' THEN organization_id = #{id}"
+    end.join("\n            ")
+  end
```
todtb, line 125: Nice! I think this is a huge first step and POC not just for blazer but for organizational security throughout the app.

app/models/user.rb
```diff
@@ -8,6 +8,17 @@ def set_default_role
     self.role ||= :user
   end
 
+  # Track the logged-in user at the model level
+  # so that we can leverage the user's organization
+  # within blazer reporting.
+  def self.current
+    Thread.current[:user]
```
haydenrou, line 15: Curious whether a gem like [request_store](https://github.com/steveklabnik/request_store) could be a good idea here. I haven't used it though, just throwing the idea out there.

app/controllers/application_controller.rb
```diff
@@ -10,4 +11,21 @@ class ApplicationController < ActionController::Base
...
   end
+
+  # Auth check for Blazer reporting
+  def require_organization
+    redirect_to root_path unless current_user && current_organization
+  end
+
+  def set_current_user
+    User.current = current_user
+  end
```
benreyn, line 22: Im curious why we need this, and we cant just use the `current_user` helper from devise directly.

app/models/user.rb
benreyn: Curious why we need to tap into `Thread` here. Is there something about just using `current_user` directly that wont work?

app/controllers/application_controller.rb
wadewinningham: The blazer `data_sources` method that's being overridden doesn't have access to controller-level variables. `current_user` isn't within it's scope.

app/models/user.rb
wadewinningham: This is a thread-safe way to keep track of the currently logged-in user at the model level, which doesn't have access to controller/view level variables like `current_user`. I just know from experience that in this context, if you don't store it on `Thread`, the value would be non-deterministic if you're running on a threaded web server.

benreyn: Ahh, got it. I see now that this is being used in an initializer. Thought on using Rails built in [ActiveSupport::CurrentAttributes](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html) for this?

wadewinningham: Our main app isn't up on Rails 6 yet, so the `CurrentAttributes` is something I wasn't familiar with. Seems like it could work. I can try it.

wadewinningham:
@benreyn @CraigJZ Just updated this to use `ActiveSupport::CurrentAttributes`.

haydenrou (approved):

benreyn:
@wadewinningham, thank you so much for making this possible! I think this is pretty close to ready. I think the last todo here is to address rubocop. I can take care of that sometime today. Thank you so much for all your work on this!

wadewinningham:
@benreyn Awesome! Thanks. Let me know if you want me to do anything else with this one. I'd considered moving some of the rake task logic into a class so it's more testable, but that depends on your priorities.

CraigJZ (approved):

PR #462, opened 2020-10-08
Unpin faker gem from master

todtb:
### Description
Un-pins faker from master branch. This was leading dependabot to open a PR every time Faker gem's master was updated. We can probably just use the latest release tag.

### Type of change

### How Has This Been Tested?
Specs still good, we weren't using anything not cut into a gem version.


mdworken (approved):
Nice catch! Really appreciate it.

PR #485, opened 2020-11-01
Update webpack-dev-server dependencies

benreyn:
Github alerted us to a high-severity security vulnerability in
node-forge, a dependency of the selfsigned package, which is a
dependency of webpack-dev-server

This commit was created by running...

```
$ yarn upgrade webpack-dev-server
```

to upgrade these dependencies and eliminate this security vulnerability


benreyn:
Worth noting, Im not sure why dependabot didn't create a PR for this automatically. We were getting these automatic PRs from dependabot for a while but they stopped coming. Perhaps we need to reconfigure dependabot for this repository?

CC: @CraigJZ

CraigJZ:
Thanks! In this case, it looks like another dependency was locking node-forge at the vulnerably version but I'm not sure why we haven't been getting other dependabot PRs. I'll take a look at the config.

CraigJZ (approved):

PR #486, opened 2020-11-01
Update a number of gems

benreyn:
We seem to not be getting this through dependabot anymore. I wonder if that is opt-in now?

cc: @CraigJZ


benreyn:
Interestingly, rubocop tagged some offenses on this branch. I added a commit to fix them, but its odd that there are offenses on the main branch 🤷

PR #507, opened 2020-12-01
[wip] Blazer rake task errors

CraigJZ:


mdworken (approved):
Sure why not 😛

benreyn:
Interestingly, this was a change I made after merging this rake task. It was a recommendation from rubocop (to use `.safe_load` over `load`) I wonder if this would have worked as is if using `.load` instead?

CraigJZ:
Ah, interesting. I tried it with .load and that does get around the alias error we encountered too. I'm still running into an issue locally with the db environment variables. I'll try to sort that out, but if it runs successfully on your local machines, I think we could merge and try it in production.

PR #523, opened 2020-12-12
Blazer rake task modifications

CraigJZ:
This branch makes a couple of modifications to the blazer setup  -

First, it updates the blazer rake task so it only creates or drops the organization roles when not run in production. This is to conform to our current hosting, which requires those roles be set manually prior to running the task.

Second, it wraps the blazer mount in devise's authenticate method to provide a more user friendly error than blazer's standard one if an unauthenticated users tries to access that route. This may not be needed once when update the blazer and report routes.

Once merged, I'll test the updated rake task on abalone-staging and make any necessary tweaks.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -38,7 +38,10 @@ Rails.application.routes.draw do
   get 'about', action: 'show', controller: 'home'
 
   mount ReportsKit::Engine, at: '/'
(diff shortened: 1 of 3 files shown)


mdworken (approved):
Looks great to me.

PR #575, opened 2021-01-07
makes seeds file idempotent

noahc:
### Description

I ran `bin/setup` and then ran `rake db:seed`, which resulted in an error being raised. I believed that the rake command had failed to populate the database correctly. Instead, it failed because the seeds had already been loaded into the database. 

The problem was that the `find_or_create` method was looking at `date_time_collected` which varied between the first run and second run. This caused the find part to never match and the create portion failed because tag had already been used and raised a uniqueness validation error. 

### Type of change

### How Has This Been Tested?

I can run `rake db:seeds` over and over again and not get any errors.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -88,27 +88,27 @@ organization_entities[:organizations].each do |org_ent|
 
     dead_animal.mortality_event = MortalityEvent.create(cohort: cohort)
 
-    male = Animal.find_or_create_by(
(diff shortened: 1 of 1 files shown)


CraigJZ (approved):

CraigJZ:
@noahc thanks for catching this. I was able to reproduce the error you described and I think the solution you've provided looks good.

Issue #535, opened 2020-12-18
Report iframe and canvas do not have a title or label


NiquiCara:
The iframe containing the graph, and the canvas displaying the graph, need an accessible title or label

![iframe-missing-label](https://user-images.githubusercontent.com/45127691/102561077-c1c8b280-40a1-11eb-9633-219688088f3a.PNG)

CraigJZ:
Closed by #580

Issue #534, opened 2020-12-18
Reports: text-box for date selector does not have an accessible label


NiquiCara:
This textbox does not have a label associated with it.

We can either add a label with `sr-only` or we can add an aria-label tag. I think either will work.

![textbox-without-label](https://user-images.githubusercontent.com/45127691/102560989-8fb75080-40a1-11eb-813e-a2297d7f854b.PNG)

CraigJZ:
Closed by #580

PR #580, opened 2021-01-09
Possible path for integrating blazer into the site navigation.

CraigJZ:
Currently, blazer is not accessible through the site's navigation. This branch offers a possible approach to that by removing the current reports routing and adding `as: reports` to the blazer routes.

An alternative approach can be seen here: https://github.com/ankane/blazer/issues/50#issuecomment-259071162. This appears to requiring forking the gem, which we have avoided doing so far.

@benreyn or @mdworken, any thoughts or preferences on this?

If the proposed option is acceptable, I can finish cleaning it up and prepare a proper PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,7 +64,7 @@ group :development do
 end
 
 # Report builder
-gem 'reports_kit'
+gem 'reports_kit' # Replaced by blazer reporting - 1/24/21
 gem 'blazer'
 
 # Excel and CSV support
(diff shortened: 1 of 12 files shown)


CraigJZ:
Just adding some notes for future reference.
- Existing reports_kit code has been annotated with the following comment `# Replaced by blazer reporting - 1/24/21`  pending future discussions. Initially, reports_kit was setup as a backup for blazer if we couldn't scope queries to an organization. With that working blazer is the primary reporting tool now.
- The nav changes made in this branch are intentionally easily reversible should the group decide to go a different direction.

Closes #534 
Closes #535

Issue #621, opened 2021-02-26
Multiple <h1>s on homepage 


metamoni:
Pages should only have one main heading (unless we decide to use multiple `<h1>s` (see #532), representing the most important idea on the page. At the moment, this is 'Abalone Analytics'.

The rest of it needs a bit of restructuring. Following [these requirements](https://www.w3.org/WAI/tutorials/page-structure/headings/) and [this example](https://www.w3.org/WAI/tutorials/page-structure/example/), we could logically structure the page like so:

```
<h2>Your Organization: White Abalone</h2>
<ul>
  <li>Animal count: 132</li>
  <li>facility count: 12</li>
  <li>cohort count: 12</li>
</ul>
```

We have a lot of `div`s on the page at the moment, all from bootstrap, so maybe jamming a list in between those isn't the best. But we can follow the same pattern and use `div`s instead of my proposed `ul` and `li` tags. As long as we keep the same styles we'll get the same result visually. But when on a screen reader, this will read out 'Abalone Analytics' as the main heading, your organization's name as `<h2>`, and then the animal/facility/cohort counts under that.

metamoni:
I'm happy to work on this one, by the way. I'll push up a PR for it :)

CraigJZ:
Thank you for adding this issue! I like your proposed structure and am okay with using a list there.

I added my thoughts on #532 but let me know if you have any concerns with that approach. Overall, my preference for this app would be using a single `<h1>` per page.

PR #622, opened 2021-02-28
add screen reader-only page title, update cards and page logo

### LINKED ISSUE #532: Background image in hero is not decorative, needs text
The hero image is set in a background, which is ok, but there is no text for it. 

We need to have "Abalone Analytics" included as a heading element within that container. Adding the `.sr-only` class should make it invisible except to screen readers.

hero-image-background

metamoni:
# Checklist:

Resolves #532 and #621

### Description
The logo in the navigation was an `h1` element. As we only want one `h1` on every page, and we want this to be the page title, this has been updated to a simple link.
An accessible title has been added to the homepage. This is only visible to screen readers, as sighted users can read the title on the hero banner. At the same time, the card titles and counts were inappropriately wrapped in `h2` and `h1` tags. To improve the structure of the page, these are now simple paragraphs under the 'Your organisation' header. (I originally considered making this a list, but since the cards are built from lots of `div`s I didn't want to have a bunch of list items wrapped in `div`s, as this would look pretty confusing. I also changed the `section` elements to `div`s, as sections should have a heading and I didn't think they were appropriate here.
Visually, there is no difference between the previous and current structure. There's a very small difference in the animal/cohort/facility count's font weight. For some reason Tailwind's `font-bold` class wouldn't work on these elements 🤔

### Type of change
Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?
In the browser, using ANDI and VoiceOver.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -26,7 +26,7 @@
   -webkit-box-shadow: -4px -1px 33px -5px rgba(0, 0, 0, 0.25);
   -moz-box-shadow: -4px -1px 33px -5px rgba(0, 0, 0, 0.25);
   box-shadow: -4px -1px 33px -5px rgba(0, 0, 0, 0.25);
-  h1 {
+  h1, .logo-text {
     font-size: 20px;
     color: #4a4a4a;
     font-weight: 500;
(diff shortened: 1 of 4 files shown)


CraigJZ (approved):
This looks great! Thank you.

I am not sure why the tailwindcss font-bold class isn't working either, but I'll take a closer look at it.

PR #670, opened 2021-05-06
Fix navigation links using tabindex

### LINKED ISSUE #533: Navigation links with drop-downs are not accessible via keyboard navigation
The three links with drop-down menus: "Uploads", "Admin", "More" are not accessible with keyboard navigation. The `tab` key skips over all three when tabbing through the menus.

We need the ability to tab into those menus, have the menus expand, and be able to select the items within. They will also need to be added to the tab index.

links-arent-accessible

librod89:
# Checklist:

Resolves #533

### Description

I added a `tabindex` attribute to each of the links in the navbar so you can use the keyboard to navigate to any of the links or dropdowns.

### Type of change

### How Has This Been Tested?

I added tests to `spec/features/home_page_statistics_spec.rb`


CraigJZ (approved):
Thanks for working on this @librod89!!

PR #742, opened 2021-07-29
Fix nil warning for tests be adding an enclosure to cohort

### LINKED ISSUE #708: Fix Warning on Cohorts Show action feature spec
The spec at `spec/features/cohorts/show_spec.rb:19` throws the following warning:

`Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead.`

Refactor the spec to clear this warning.

FionaDL:
Resolves #708 

 (fix or feature that would cause existing functionality to not 
work as expected)
* Documentation update

### How Has This Been Tested?

This is a fix for the test in feature/cohorts/show_spec . Before the tests were complaining because cohort.enclosure&.name was nil, so there was no point in checking. I added an enclosure to the cohort so that we could check if the name shows up on the page.

### DIFF
--- a/spec/features/cohorts/show_spec.rb
+++ b/spec/features/cohorts/show_spec.rb
@@ -9,6 +9,10 @@ describe "When I visit the cohort Show page" do
 
   it "Then I see information of a specific cohort" do
     cohort = create(:cohort)
+    enclosure = create(:enclosure)
+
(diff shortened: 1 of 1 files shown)


CraigJZ (approved):
@FionaDL This looks good. Thank you!!

PR #820, opened 2021-09-28
#540 add aria-label on links

### LINKED ISSUE #540: Enclosures: hidden action links lack names
These links don't display, but they still exist. They don't have names, but they still show up in tab order.

We should probably not display them at all if they are not supposed to be used. If we need them to be there, they should not be focusable if they do not need to be used. If they must be focusable, they need to have names.

enclosures-hidden-links

vurtn:
Sorry for my english.

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas,
- [ ] Title include "WIP" if work is in progress.

Resolves #540 

### Description
I added an aria-label attribute on links to not overload the view for the common use. I don't have problem with icons on firefox and brave web browsers.

### Type of change

### How Has This Been Tested?

I used the same tool to check accessibility which is ANDI. The alert message is no longer there. 

### Screenshots
screen540-2
screen540-1


CraigJZ (approved):
Thank you!

PR #827, opened 2021-10-02
refactor animals table

metamoni:
Resolves #796 

### Description
Rename collection-related columns in `animals` table to more generic terms, and clarify the status of an animal (spawned or collected). This PR includes the necessary migrations to:
- rename `collection_year` to `entry_year`, `date_time_collected` to `entry_date`, `collection_position` to `entry point`
- add `""` as default value for `entry_point`
- add a `collected` column, which is `true` for collected and `false` for spawned animals

I have also updated all references to the renamed columns, as well as:
- the permitted params in the `animals_controller`
- the Animals index & show views, and the form
- the Animal model with new validation to require a non-blank entry point for collected animals
- seeds, specs, factories

### Type of change
(I think this is a new feature?)
* New feature (non-breaking change which adds functionality)

### How Has This Been Tested?
- updated existing specs
- added specs to test the new validation in the Animal model

### Screenshots
_Animals table_
Screenshot 2021-10-02 at 22 16 06

_Spawned animal_ 
Screenshot 2021-10-02 at 22 16 26

_Collected animal_
Screenshot 2021-10-02 at 22 16 36

PR #829, opened 2021-10-03
add collected filed to animal form

### LINKED ISSUE #828: Add "collected" field to add/edit Animal form
As part of #796 we're adding a `collected` field to the Animals table. When this is `true`, we need to provide a non-null `entry_point`, or else we'll get a validation error. When it is `false`, the animal was spawned and an `entry_point` field is not needed (this will be automatically populated with an empty string in the database).

Once [this PR](https://github.com/rubyforgood/abalone/pull/827) is merged, add a "Collected" field with two radio options - "Yes" and "No". If the user selects "Yes", we should display a required "Entry Point" field where the user will enter the location.

Add changes in `app/views/animals/_form.html.erb`. The new "Collected" and "Entry Point" fields could be added under the "Entry Date" field.

Screenshot 2021-10-02 at 22 36 55

metamoni:
Resolves #828 

This PR should only be merged after https://github.com/rubyforgood/abalone/pull/827 is merged.

### Description
[This PR](https://github.com/rubyforgood/abalone/pull/827) introduces a `collected` column in the Animals table, which needs to be added to the New/Edit Animal form. When its value is `true`, we also need to display an Entry Point field.

### Type of change

### How Has This Been Tested?
Manually (we don't have a testing framework for Stimulus yet)

### Screenshots
_New/Edit Animal forms_
https://user-images.githubusercontent.com/22390758/135749461-aa3e20af-c932-488f-8159-cc13ecce797d.mov

_Submitting a new animal_
https://user-images.githubusercontent.com/22390758/135749483-5668133a-4f23-4b0a-b9bc-aae8e835fdee.mov


CraigJZ:
This looks great. I like the approach you used with the stimulus controllers.
Thank you!!

PR #837, opened 2021-10-08
Resolves issue #724: Org-scoped Blazer queries

### LINKED ISSUE #724: Only display saved queries in blazer for the current organization
The blazer interface displays links to all saved queries for all organizations. Queries for organizations other than the one a user is signed in for can not be accessed and throw an error when clicked. We should only display saved queries for the current organization. One potential path would be to set row level security on the blazer tables likely scoped to the creator's organization.

mattwd7:
Resolves #724 

Introduces an app/overrides directory for overriding engine controllers (and models)
in the conventional Rails way with class_evals loaded on application start.
(read more at https://edgeguides.rubyonrails.org/engines.html#overriding-models-and-controllers).
Override is applied to the Blazer queries_controller method :set_queries, scoping returned
queries on :home and :index actions to only those belonging to the current_user's
organization.

This change depends upon Blazer version 2.4.7, which I've now locked in the Gemfile.

This solution feels fragile on account of its reliance on the entire :set_queries method defined by the Blazer queries_controller. A change to this logic in a future version of Blazer will cause problems on gem update. A better way to override would be to add a controller method :base_query or :scoped_query to the Blazer gem in this controller for more finite overriding of this smaller method, e.g.

[# blazer/queries_controller.rb](https://github.com/ankane/blazer/blob/master/app/controllers/blazer/queries_controller.rb)
```
def set queries
  - @queries = Blazer::Query.named.select(:id, :name, :creator_id, :statement)
  + @queries = scoped_query.named.select(:id, :name, :creator_id, :statement)
  ...
end

+ def scoped_query
+   Blazer::Query
+ end
```

### Type of change

### How Has This Been Tested?
RSpec tests added to support this change.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -66,6 +66,7 @@ AllCops:
     - 'db/**/*'
     - 'bin/**/*'
     - 'config/routes.rb'
+    - 'app/overrides/controllers/blazer/queries_controller_override.rb'
   # Default formatter will be used if no `-f/--format` option is given.
   DefaultFormatter: progress
   # Cop names are displayed in offense messages by default. Change behavior
--- a/Gemfile
+++ b/Gemfile
@@ -67,7 +67,11 @@ end
 
 # Report builder
 gem 'reports_kit' # Replaced by blazer reporting - 1/24/21
-gem 'blazer'
+# Dependency on Blazer version 2.4.7 due to app-specific overrides living in
+# app/overrides that depend on the Blazer code at this version of the gem. Updating
+# this gem will require review of our overridden classes with changes to code in the
+# desired Blazer gem version.
+gem 'blazer', '2.4.7'
 
 # Excel and CSV support
 gem 'creek'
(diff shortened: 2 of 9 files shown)


CraigJZ:
@mattwd7 Thank you for working on this. This looks like a good solution. I agree that it does feel fragile, but I think I prefer this path to forking the gem. Perhaps we could add some documentation about the dependency on 2.4.7. We could add a comment in the gemfile noting this and add a comment about this to [Blazer Reporting Setup Instructions](https://github.com/rubyforgood/abalone/blob/main/blazer_reporting.md).

I enabled the CI - my apologies for the delay on this. It looks like there are some rubocop violations on the `queries_controller_override.rb` file. I suspect we won't want to fix these because this file needs to match up with the gem. If that's the case, lets exclude the file in the `.rubocop.yml` file.

I think we should be good to merge this after those changes. Thanks again!

PR #904, opened 2021-11-11
Name validation

### LINKED ISSUE #437: Add model level validations to Cohort

LawrenceWhalen:
Resolves #437 

### Description
Added validations to cohort model for the `:name` property. Model now test for presence of ':name', and will fail out actions that do not set or remove the attribute. ':name' was also set to be unique for the organization the cohort was created for. Multiple organizations can create cohorts with the same ':name', but the ':name' cannot be repeated for that orginization.  

### Type of change

### How Has This Been Tested?

Added validation tests to `cohort_spec.rb`. Used standard shoulda-matcher tests for `:name` presence. 
Shoulda-matcher does not test for uniqueness of attributes scoped to other models, so tests were written ensuring that an instance became invalid if it was a repeated `:name`. A second expectation tests that a `:name` was not locked for other organizations. Both were tested using the boolean `.valid?` method.

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -13,6 +13,8 @@ class Cohort < ApplicationRecord
 
   belongs_to :enclosure, required: false
 
+  validates :name, presence: true, uniqueness: { scope: :organization }
+
   delegate :name, to: :enclosure, prefix: true, allow_nil: true
 
   def self.exportable_columns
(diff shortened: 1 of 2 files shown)


CraigJZ (changes requested):
@LawrenceWhalen This looks good! Thank you!
It looks like there are 2 minor rubocop offenses. Once those are fixed we can go ahead and merge this.

LawrenceWhalen:
@CraigJZ Variable names changed to normalcase.

CraigJZ (approved):
Thank you!!

PR #924, opened 2021-12-14
Resolves #901: Prevent the deletion of Measurement Types that have already been used in measurements

### LINKED ISSUE #901: Prevent the deletion of Measurement Types that have already been used in measurements
We would like to prevent the deletion of measurement types that are already being used in measurements. The way we handle this for Exit Types can serve as an example for handling this issue. 

- [ ] Update the controller destroy action to prevent deletion of measurement types that have associated measurements and provide a helpful flash notice 
- [ ] Update the measurement types index to hide the delete icon for measurement types that have associated measurements
- [ ] Add specs to test both of these changes
- [ ] Update any existing relevant specs

mattwd7:
Resolves #901  

### Description
With this change, measurement_types cannot be deleted if they have any dependent measurement records. In writing this code, I discovered ActiveRecord's :restrict_with_error association dependency on the model which prevents destruction at a very low level. Issues and errors resulting from a failed attempt to destroy a measurement_type are added to the object's errors for display to users and engineers working with these records. (Sidebar: we may want to circle back to exit_types and apply this same pattern for safer handling of those records outside the Controller. )

### Type of change
New feature / Bug fix

### How Has This Been Tested?
Rspec tests have been added for measurement_type model, controller, and view

### DIFF
--- a/app/controllers/measurement_types_controller.rb
+++ b/app/controllers/measurement_types_controller.rb
@@ -30,8 +30,11 @@ class MeasurementTypesController < ApplicationController
   end
 
   def destroy
-    @measurement_type.destroy
-    redirect_to measurement_types_url, notice: 'Measurement Type was successfully destroyed.'
+    if @measurement_type.destroy
+      redirect_to measurement_types_url, notice: 'Measurement Type was successfully destroyed.'
+    else
+      redirect_to measurement_types_url, alert: @measurement_type.errors.full_messages.join(". ")
+    end
   end
 
   def edit; end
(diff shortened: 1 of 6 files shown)


mattwd7:
image

mattwd7:
Ready for review, @CraigJZ !

CraigJZ (approved):
Thanks @mattwd7 and apologies for the delay in reviewing this.
Looks great and I do like the pattern used here with `:restrict_with_error`. I'd be open to a refactor of exit_types to use this same pattern.

PR #926, opened 2021-12-15
Resolves #899: Provide Error Message for Invalid Measurement Types in import

### LINKED ISSUE #899: Provide Error Message for Invalid Measurement Types in import
Measurement Types can be defined at the organizational level. We want to provide a helpful error message when a measurement is submitted with a measurement_type that is not defined for the organization. Currently, the measurement upload hangs if a measurement_type that isn’t available in an organization is used. The way we handle a missing or invalid subject_type might serve as an example for handling this issue. 

Screen Shot 2021-11-07 at 1 03 29 PM

- [ ] Update the measurement upload provide the user with an error message when a measurement can’t be processed because the measurement_type is not available 
- [ ] Add specs to cover this scenario

mattwd7:
Resolves #899

### Description
This change is necessarily larger than the issue description might imply.
The importing tools were already setup to provide an error message to the user when attempting to import a non-existent measurement_type for a new Measurement record. However, the import tooling had hard-coded a set of allowable measurement_types, likely at a time before Users could create their own custom measurement_types.

This change enables error messaging to Users by providing the following:
* Removes hard-coding of measurement_types
* Decouples CsvImporter from Measurements and MortalityEvents
* Separates the upload of Measurements from MortalityEvents (which now have distinct dropdown options at the ImportFile page)
* Additional expectation was added to an existing spec example in `spec/lib/csv_importer_spec.rb` to confirm that this change resolves issue #899 

### Type of change

### How Has This Been Tested?

New and existing rspec tests, as well as upload of Measurements and MortalityEvents running a local server.

### DIFF
--- a/app/controllers/file_uploads_controller.rb
+++ b/app/controllers/file_uploads_controller.rb
@@ -9,10 +9,6 @@ class FileUploadsController < ApplicationController
   # and process the data with a MeasurementJob.
   before_action :set_processed_file, only: %i[show destroy]
 
-  FILE_UPLOAD_CATEGORIES = CsvImporter::CATEGORIES.map do |category|
-    [category, category.delete(' ')]
-  end.freeze
-
   def index
     @pagy, @processed_files = pagy(ProcessedFile.for_organization(current_organization).order(updated_at: :desc))
     @processed_files.each { |processed_file| authorize! :index, processed_file }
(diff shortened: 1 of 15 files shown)


mattwd7:
Let me know what I can do to make brakeman and rubocop happy

mattwd7:
Also, I think we can apply this same pattern to Animals, Cohorts, and Enclosures, putting all uploads in the same place!

CraigJZ:
@mattwd7 I'm so sorry for the delay in reviewing this one. With the holidays, things have been a bit hectic. I'll give it a more thorough look first thing on Monday and follow up. Thanks for taking this one on! 

I really like the idea of decoupling the CSVImporter from the individual models as well as potentially using the same pattern for the other models. This would be a nice improvement to one of the more confusing parts of the app.

One reason for some of the existing patterns is to allow users to add mortality events as part of their measurement uploads. In most cases, they currently collect all of that data in single csv files. The sample file [here](https://github.com/rubyforgood/abalone/blob/main/public/samples/animal_count_measurements_template.csv) offers an example. We'd definitely be open to simplifying this as proposed, though, as long as we can minimize the impact on users.

mattwd7:
Sounds good, @CraigJZ ! I've updated to avoid the Brakeman issue and I've ignored the blocking Metrics/ModuleLength Cop in import_job.rb, at least for now... (I took a stab at refactoring into smaller pieces to appease Rubocop, but it was not so straight-forward and feels like it's own, separate effort if it's determined to be a priority). Ready for review!

mattwd7:
bump @CraigJZ !

CraigJZ:
@mattwd7 Thanks for the nudge and I'm sorry again for the delay in finishing up the review of this. Thanks for taking care of the brakeman issue. Ignoring the blocking cop sounds good to me. I agree that a refactor feel like a separate task. After we merge this, I'll make a task for that but I don't think it is a high priority.

I've set aside some time this weekend to test this out with some sample user data and get this wrapped up. Again, I'm really sorry for the delay on this one and appreciate you taking it on.

CraigJZ (approved):

CraigJZ:
@mattwd7 Looks great! Thanks again!!

PR #935, opened 2021-12-22
Move exit_type destroy restriction from controller to model

mattwd7:
Follow-through related to #901 

### Description
This commit moves the ExitType destroy restriction from the exit_types_controller to the model.

### Type of change
Refactor

### How Has This Been Tested?

Additional spec examples added to ExitType model

### DIFF
--- a/spec/models/exit_type_spec.rb
+++ b/spec/models/exit_type_spec.rb
@@ -13,5 +13,24 @@ RSpec.describe ExitType, type: :model do
     it_behaves_like OrganizationScope
 
     it { is_expected.to validate_presence_of(:name) }
+
+    context "with an associated mortality_events" do
+      subject(:exit_type) { create(:exit_type) }
+
+      before do
+        create(:mortality_event, exit_type: exit_type)
+      end
+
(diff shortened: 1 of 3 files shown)


spec/models/exit_type_spec.rb
```diff
@@ -13,5 +13,24 @@
...
+    context "with an associated mortality_events" do
+      subject(:exit_type) { create(:exit_type) }
+
+      before do
+        create(:mortality_event, exit_type: exit_type)
+      end
+
+      it "cannot be destroyed" do
+        expect { exit_type.destroy }
+          .not_to change(MeasurementType, :count)
```
CraigJZ, line 26: Just one minor change. I think this spec should check that the ExitType count has not changed.

CraigJZ (changes requested):
@mattwd7 Thanks for this refactor! I think we just need one small change in the Exit Type specs.

mattwd7:
Oh oops! I'll update

CraigJZ (approved):
Thanks for making that update. Looks great!!

PR #1119, opened 2023-07-27
only allow admins to edit their own organization's measurements

### LINKED ISSUE #1029: Admin should only edit their own organization's measurements
Admin for an organization should only be able to edit their own organization's measurements. Currently, an admin can edit all organization's measurements. This will be partially handled by #1028 but we should also limit authorization for these actions programatically.

### Criteria for completion

- [ ] Restrict editing measurements to the organization level. Editing is already limited to admin users. Those admin should only be able to edit their own organization's measurements.
- [ ] Add specs to exercise this new behavior.

metamoni:
Resolves #1029

### Description
Admin users are currently able to edit any organization's measurements. We want to limit this to _their own organization_. This PR introduces a new `current_org_measurement?` check and redirects the user to the measurements table if the check fails.

### Type of change

### How Has This Been Tested?
Updated request specs with further examples.

### DIFF
--- a/app/controllers/measurements_controller.rb
+++ b/app/controllers/measurements_controller.rb
@@ -8,7 +8,9 @@ class MeasurementsController < ApplicationController
 
   def show; end
 
-  def edit; end
+  def edit
+    redirect_to measurements_path, alert: 'Not authorized to edit this measurement' unless current_org_measurement?
+  end
 
   def update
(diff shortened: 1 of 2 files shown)


metamoni:
The CI is failing [with a weird error](https://github.com/rubyforgood/abalone/actions/runs/5678000009/job/15387668708), not sure what's causing it

